PR #21, opened 2018-05-09
Add default props on buy-button component

brunojdo:
Add default props, passing `salesChannel` prop that was never used.

#### Types of changes
- [x] Breaking change (fix or feature that would cause existing functionality to change)


packages/buy-button/src/index.js
```diff
@@ -11,12 +11,18 @@ import orderFormQuery from './queries/orderFormQuery.gql'
  * BuyButton Component. Adds a list of items to the cart.
  */
 export class BuyButton extends Component {
+  static defaultProps = {
+    quantity: 1,
+    seller: 1,
+    salesChannel: 1,
```
lucasecdb: Same as `seller`

packages/buy-button/src/index.js
```diff
@@ -11,12 +11,18 @@ import orderFormQuery from './queries/orderFormQuery.gql'
  * BuyButton Component. Adds a list of items to the cart.
  */
 export class BuyButton extends Component {
+  static defaultProps = {
+    quantity: 1,
+    seller: 1,
```
lucasecdb: The `seller` is defined as a string in the propTypes

packages/buy-button/src/index.js
andreldsa: Agreed

PR #23, opened 2018-05-09
Add initial version of shipping simulator

lucasecdb:
Add the shipping simulator component.

The component wasn't implemented yet.

N/A

screenshot-2018-5-9 react app 2

#### Types of changes
- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)


packages/product-details/src/FreightSimulator.js
```diff
@@ -0,0 +1,66 @@
...
+
+  handleClick = e => {
+    e.preventDefault()
+
+    // TODO: call graphql
+  }
+
+  formatMessage = id => {
+    return this.props.intl.formatMessage({ id })
+  }
```
estacioneto: :pick: No need. Why don't you use the `<FormattedMessage />` ?

lucasecdb: Because the `<FormattedMessage />` puts an additional `span` tag, and I want the text to be a direct children of `label` for accessibility purpose

PR #73, opened 2018-05-30
 Add isOneClickBuy attribute to the BuyComponent 

salesfelipe:
Add isOneClickBuy attribute to the BuyComponent 

**Product-Summary**: https://github.com/vtex-apps/product-summary/pull/39

BuyButton has the option of redirecting the user straight to checkout after being clicked.

[workspace](https://oneclickbuy--storecomponents.myvtex.com/)

#### Types of changes
- [X] New feature (a non-breaking change which adds functionality)


react/components/BuyButton/index.js
```diff
@@ -102,17 +105,23 @@ export class BuyButton extends Component {
...
   }
 }
 
+BuyButton.defaultProps = {
+  isOneClickBuy: false,
+  quantity: 1,
+  seller: 1,
+}
+
 BuyButton.propTypes = {
```
lucasecdb: Move this and the default props to static attribute

salesfelipe: Why?

salesfelipe: This way the code is easier to read and looks less messy!

lucasecdb: what's your metric to 'easier'?

salesfelipe: I mean that this isn't really a part of the code that we will be changing a lot or affects the functionality, the way I see it is when putting the propTypes inside the class you gonna make a lot bigger and harder to read, therefore removing the focus from what really matters, which is the render function and the others functions that controls state and events.

salesfelipe: I rather put it in another file then inside of the class it self.

react/components/BuyButton/index.js
```diff
@@ -8,30 +8,28 @@ import { injectIntl, intlShape } from 'react-intl'
 import ADD_TO_CART_MUTATION from './mutations/addToCartMutation.gql'
 import ORDER_FORM_QUERY from './queries/orderFormQuery.gql'
 
-const EVENT_SUCCESS = 'item:add'
-const EVENT_ERROR = 'message:error'
+const CONSTANTS = {
+  EVENT_SUCCESS: 'item:add',
+  EVENT_ERROR: 'message:error',
+  ERROR_MESSAGE_ID: 'buybutton.add-failure',
+  CHECKOUT_URL: '/checkout/#/cart',
```
brunojdo: I don't think that's a responsibility of buy-button make the redirect. He should send the event that someone click and the component who have to decide if redirect or not.

react/components/BuyButton/index.js
lucasecdb: yeah, but I only really asked to change because of the other components, for consistency

react/components/BuyButton/index.js
andreldsa: But, we don't have an event center that handles incoming events to decide what action to make.

PR #107, opened 2018-07-08
Hide fixed header while in edit mode

firstdoit:
In edit mode, fixed elements shouldn't be rendered, except when on full-screen (which effectively disables edit mode)

Superposition of top menu and admin controls.

#### Types of changes


react/components/Header/components/Modal.js
```diff
@@ -8,8 +8,8 @@ const getModalRoot = () => {
   if (typeof _modalRoot === 'undefined') {
     _modalRoot = document.createElement('div')
     _modalRoot.classList.add('vtex-modal-root')
-
-    document.body.appendChild(_modalRoot)
+    const container = document.getElementsByClassName('render-container')[0]
```
lucasecdb: Use `document.querySelector('.render-container')` :pick:

PR #136, opened 2018-08-13
Refact Content Loaders to work on Firefox

estacioneto:
On Firefox, it was not possible to determine the SVG's style via CSS/Style prop. So, what works on Chrome and Safari, doesn't work on Firefox. To work in all browsers, the found solution was pass every style via elements properties.
In `ProductPrice`, for example, we have:
```jsx
if ((showListPrice && isNil(listPrice)) || isNil(sellingPrice)) {
   return <Price.Loader {...this.props.styles} />
}
```

Inside `Price.Loader`:
```jsx
<rect
   height="0.75em"
   width="50%"
   x="25%"
   {...loaderProps['vtex-price-list__container--loader']}
   />
```

Therefore, components like `ProductDetails` renders `ProductPrice` this way:
```jsx
<ProductPrice
  styles={productPriceLoaderStyles}
  listPrice={path(['ListPrice'], this.commertialOffer)}
  {/* Price props */}
  />
```

And:
```jsx
const productPriceLoaderStyles = {
  'vtex-price-list__container--loader': {
    x: 0,
    width: '7.219em',
    height: '0.56em',
  },
  /* ... */
}
```

Summing up, what was done via CSS classes, now will be done via JSX.

Content Loaders not rendering on Firefox.

[Access the workspace](https://estacio--storecomponents.myvtex.com/)

image

#### Types of changes

### DIFF
--- a/react/components/ProductImages/index.js
+++ b/react/components/ProductImages/index.js
@@ -43,7 +43,6 @@ class ProductImages extends Component {
       'vtex-product-image__selected-image--loader': {
         x: '10%',
         y: '5%',
-        'max-width': '19.782em',
         width: '80%',
         height: '75%',
         ...loaderProps['vtex-product-image__selected-image--loader'],
@@ -51,7 +50,6 @@ class ProductImages extends Component {
       'vtex-product-image__thumbnail-slider--loader': {
         x: '10%',
         y: '85%',
-        'max-width': '19.782em',
         width: '80%',
         height: '10%',
         ...loaderProps['vtex-product-image__thumbnail-slider--loader'],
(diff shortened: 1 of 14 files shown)


react/components/ProductImages/index.js
```diff
@@ -38,8 +38,42 @@ class ProductImages extends Component {
...
   }
 
-  static Loader = props =>
-    props.isVertical ? (
+  static Loader = (loaderProps = {}) => {
+    const styles = {
+      'vtex-product-image__selected-image--loader': {
+        x: '10%',
+        y: '5%',
+        'max-width': '19.782em',
```
lucasecdb: `maxWidth`

lucasecdb: Also, use the `rem` unit, not `em`

estacioneto: Actually, this property doesn't work. It's not necessary.

PR #196, opened 2018-11-19
Fix SVG warnings

rerissondaniel:

Replace the unit in which the width and height of content loaders were defined.

The height and width props of the [react-content-loader](https://github.com/danilowoz/react-content-loader) were being defined using percentage but as these values are used to define the viewBox of the content loader warnings were shown. This PR replace these percentage values with user units, adequate to each content loader. 

[Access this workspace](https://svgerrorpd--storecomponents.myvtex.com/iphone-xs/p), open the console. Select slow 3g in the network tab so you can see the content loaders.

#### Types of changes
- [X] Bug fix (a non-breaking change which fixes an issue)


CHANGELOG.md
```diff
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
```
luciannojunior: Remove 9th line

PR #220, opened 2018-12-04
Make Logo properly receive width and height properties.

matheusps:
#### What is the purpose of this pull request? What problem is this solving?
As the title says.

#### How should this be manually tested?
[Workspace](https://logosize--storecomponents.myvtex.com/)

#### Types of changes
- [x] Bug fix (a non-breaking change which fixes an issue)
- [ ] New feature (a non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Requires change to documentation, which has been updated accordingly.

PR #223, opened 2018-12-04
Feature/logo onload

lbebber:
Adds `onLoad` event to Logo component, to be able to get its size.

### DIFF
--- a/react/components/Logo/index.js
+++ b/react/components/Logo/index.js
@@ -18,6 +18,8 @@ export default class Logo extends Component {
     height: PropTypes.number,
     /** Set label visibility */
     showLabel: PropTypes.bool,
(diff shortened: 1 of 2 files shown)


react/components/Logo/index.js
```diff
@@ -26,19 +28,51 @@ export default class Logo extends Component {
     showLabel: true,
     width: 493,
     height: 177,
+    onLoad: () => {},
```
kevinch: Useful?

lbebber: default props, pra nao quebrar se nao existir

lbebber:
I don't think `onLoad` is needed anymore, and the sizing should be fixed by https://github.com/vtex-apps/store-components/pull/220

PR #229, opened 2018-12-16
Fix Shipping Simulator Price Table Number Format

juliomoreira:
Converts price value to decimal instead of cents to fix price display.

shipping-fix

### DIFF
--- a/react/components/ShippingSimulator/components/ShippingTableRow.js
+++ b/react/components/ShippingSimulator/components/ShippingTableRow.js
@@ -25,7 +25,7 @@ const ShippingTableRow = ({ name, shippingEstimate, price, intl }) => {
   } else if (price === 0) {
     valueText = intl.formatMessage({ id: 'shipping.free' })
   } else {
-    valueText = intl.formatNumber(price, { style: 'currency', currency: 'BRL' })
+    valueText = intl.formatNumber(price/100, { style: 'currency', currency: 'BRL' })
   }
 
   return (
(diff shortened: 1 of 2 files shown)


react/components/ShippingSimulator/components/ShippingTableRow.js
```diff
@@ -25,7 +25,7 @@ const ShippingTableRow = ({ name, shippingEstimate, price, intl }) => {
   } else if (price === 0) {
     valueText = intl.formatMessage({ id: 'shipping.free' })
   } else {
-    valueText = intl.formatNumber(price, { style: 'currency', currency: 'BRL' })
+    valueText = intl.formatNumber(price/100, { style: 'currency', currency: 'BRL' })
```
matheusps: I think this division should be optional.

juliomoreira: I don't see why.

The price value comes directly from the getShippingEstimates query which returns the value in cents, and it's totally fine, should be an standard if it's not already.

React-intl formatNumber method doesn't care about it and it's not part of his job to manipulate the values of attributes but it's expecting a decimal based number to be formatted.

So I don't see why we should control if the division is done or not, only if you're planing to change returned output of the query.

matheusps: Good point, it makes sense. Just add a CHANGELOG and your PR will be good to go, for me.

matheusps: Yeah, good argument. It makes sense. For me, just update the CHANGELOG file with your changes and the PR is good to go. 🙂

juliomoreira: I'll do that.

Sorry for not going through all the PR guidelines.

matheusps: No problem 🙂

PR #259, opened 2019-01-17
Add CSS Modules in ProductPrice

brunojdo:
As the title says

https://css--storecomponents.myvtex.com/st-tropez-shorts/p
#### Types of changes


react/components/ProductPrice/index.js
```diff
@@ -128,7 +129,7 @@ class Price extends Component {
 }
 
 Price.Loader = (loaderProps = {}) => (
-  <div className={classNames('vtex-price vtex-price-loader', loaderProps.loaderClass)}>
+  <div className={classNames(`${productPrice.priceContainer} ${productPrice.priceContainerLoader}`, loaderProps.loaderClass)}>
```
Klynger: Instead of using string interpolation inside classNames you can just
```javascript
classNames(productPrice.priceContainer, productPrice.priceContainerLoader, loaderProps.loaderClass)
```

react/components/ProductName/index.js
```diff
@@ -100,32 +100,16 @@ class ProductName extends Component {
...
-    })
-
-    const productReferenceClasses = classNames(`${productName.productReference}`, {
-      [`${productReferenceClass}`]: productReferenceClass
-    })
-
     return (
-      <Wrapper className={productNameClasses}>
-        <span className={productBrandClasses}>
+      <Wrapper className={`${productName.productNameContainer}`}>
```
theomoura: You don't need string interpolation here. Just use `className={productName.productNameContainer}`, same as above cases

react/components/ProductPrice/index.js
```diff
@@ -74,26 +75,26 @@ class Price extends Component {
     const differentPrices = showListPrice && sellingPrice !== listPrice
 
     return (
-      <div className={classNames('vtex-price', className)}>
+      <div className={classNames(`${productPrice.priceContainer}`, className)}>
```
theomoura: The same here, no need for string interpolation

react/components/ProductPrice/index.js
```diff
@@ -128,7 +129,7 @@ class Price extends Component {
 }
 
 Price.Loader = (loaderProps = {}) => (
-  <div className={classNames('vtex-price vtex-price-loader', loaderProps.loaderClass)}>
+  <div className={classNames(productPrice.priceContainer, productPrice.priceContainerLoader, loaderProps.loaderClass)}>
```
Klynger: You're missing `productPrice.priceContainerLoader`

react/components/ProductName/index.js
```diff
@@ -48,7 +48,7 @@ class ProductName extends Component {
   }
 
   static Loader = (loaderProps = {}) => (
-    <div className={`${productName.productNameContainer} ${productName.productNameLoader} ${loaderProps.className}`}>
+    <div className={productName.productNameContainer, productName.productNameLoader, loaderProps.className}>
```
Klynger: You're missing the usage of `classnames` funciton here, this code is only applying the classes of `loaderProps.className`

react/components/ProductPrice/index.js
brunojdo: I don't understand, the handle is defined

PR #263, opened 2019-01-22
Support to CSS Modules in SKUSelector

GustavoMA:

As described in the title.

[https://cssmodulesskuselector--storecomponents.myvtex.com](https://cssmodulesskuselector--storecomponents.myvtex.com)

#### Types of changes


react/components/SKUSelector/components/SKUSelector.js
```diff
@@ -3,13 +3,15 @@ import React, { Component } from 'react'
...
+import styles from '../styles.css'
+
 /** Renders the main and the secondary variation, if it exists. */
 export default class SKUSelector extends Component {
   render() {
     const { selectedId, onSelectSku, mainVariation, secondaryVariation, maxSkuPrice } = this.props
 
     return (
-      <div className="vtex-sku-selector">
+      <div className={`${styles.skuSelectorContainer}`}>
```
JuanAlmeida12: You can put only `className={styles.skuSelectorContainer}` instead of interpolation string.

react/components/SKUSelector/components/SelectorItem.js
```diff
@@ -22,14 +24,14 @@ export default class SelectorItem extends PureComponent {
     return (
       <div
         className={classNames(
-          'vtex-sku-selector__item relative di pointer flex items-center',
-          isImage && 'vtex-sku-selector__item-image'
+          `${styles.skuSelectorItem} relative di pointer flex items-center`,
+          isImage && `${styles.skuSelectorItemImage}`
```
JuanAlmeida12: Here too

Klynger: In this case you should actually do something like this
```javascript
className={classNames(
          `${styles.skuSelectorItem} relative di pointer flex items-center`,
           {
             [styles.skuSelectorItemImage]: isImage
           })
```

react/components/SKUSelector/components/SelectorItem.js
```diff
@@ -22,14 +24,14 @@ export default class SelectorItem extends PureComponent {
...
         )}
         onClick={onClick}
       >
         <div
           className={classNames(
-            'absolute frame-around b--action-primary br3 bw1',
+            `absolute ${styles.frameAround} b--action-primary br3 bw1`,
             isSelected && 'ba'
```
Klynger: Same

react/components/SKUSelector/components/SelectorItem.js
```diff
@@ -42,7 +44,7 @@ export default class SelectorItem extends PureComponent {
           <div
             className={classNames(
               'absolute absolute--fill',
-              isAvailable || 'diagonal-cross'
+              isAvailable || `${styles.diagonalCross}`
```
Klynger: This seems weird to me, what happens when isAvailable is true? This expression is just returning a boolean then, right?

GustavoMA: When it's a `||` it returns the last truty value (from left to right).
However, I think it's better to fix and use the classNames syntax with objects.

Klynger: Yeah, but `isAvailable` is a boolean, so when its value is truthy it will return it, so it will return `true`, but you're changing it, so whatever :100:

GustavoMA: Yep, It's a Boolen, if it is `true` it will return the value in `styles.diagonalCross` if it is truthy, if it is falsy will return `true`. I think someone confused `||` with `&&` when made this line of code.

GustavoMA: Wow, I'm totally wrong. `||` operator returns the first truthy value.

PR #264, opened 2019-01-22
Added props that allow search icon to be on left or right in SearchBar.

matheusps:
#### What is the purpose of this pull request? | What problem is this solving?
As the title says.

This [workspace](https://restyleheadermobile--storecomponents.myvtex.com) on mobile mode.

#### Types of changes

### DIFF
--- a/react/components/SearchBar/components/AutocompleteInput.js
+++ b/react/components/SearchBar/components/AutocompleteInput.js
@@ -2,7 +2,8 @@ import React, { Component } from 'react'
 import PropTypes from 'prop-types'
 import classNames from 'classnames'
 
-import { Input, IconClose, IconSearch } from 'vtex.styleguide'
+import { Input, IconClose } from 'vtex.styleguide'
+import Icon from 'vtex.use-svg/Icon'
 
 import styles from '../styles.css'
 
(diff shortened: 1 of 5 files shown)


react/components/SearchBar/components/SearchBar.js
```diff
@@ -30,11 +32,13 @@ export default class SearchBar extends Component {
...
         onGoToSearchPage={onGoToSearchPage}
         inputValue={inputValue}
+        hasIconLeft={hasIconLeft}
+        iconClasses={iconClasses}
       />
     )
 
     const mainClasses = classNames(
-      `${styles.searchBarContainer} w-100 ph5 ph0-ns pb5 pb0-ns`
+      `${styles.searchBarContainer}`
```
JuanAlmeida12: Remove interpolation string use only `styles.searchBarContainer`

react/components/SearchBar/components/AutocompleteInput.js
```diff
@@ -26,17 +27,21 @@ export default class AutocompleteInput extends Component {
   }
 
   render() {
-    const { onGoToSearchPage, onClearInput, compactMode, value, ...restProps } = this.props
+    const { onGoToSearchPage, onClearInput, compactMode, value, hasIconLeft, iconClasses, ...restProps } = this.props
 
     const suffix = (
       <span className="flex items-center pointer" onClick={value ? onClearInput : undefined} >
```
Klynger: This `onClick` shouldn't just be `onClick={value && onClearInput}`?

react/components/SearchBar/components/AutocompleteInput.js
```diff
@@ -26,17 +27,21 @@ export default class AutocompleteInput extends Component {
...
         {value
           ? <IconClose className="pa0" size={20} />
-          : <IconSearch />
+          : !hasIconLeft && <Icon id="hpa-search" className={iconClasses} />
```
Klynger: I think is better to use a getter here instead of using a ternary operator and then an expression. Something like

```javascript
get myIcon() {
  if (value) {
    return <IconClose ... />
  } else {
    return !hasIconLeft && ...
  }
}
```

Klynger: About the name of the getter I don't know a good one, in `product-summary` we was doing some getters like `get renderProductName()`, `get renderProductPrice()` and so on. But they all return the same component, `get renderIcon()` maybe a good name.

react/components/SearchBar/components/AutocompleteInput.js
matheusps: Yes, it should.

PR #265, opened 2019-01-22
Feature/css modules categories highlights

paul0vinicius:
Use CSS Modules on CategoriesHighlights

The use of global.css

[Workspace link](https://cssmodulescategorieshighlights--storecomponents.myvtex.com/)

#### Types of changes


react/components/CategoriesHighlights/components/CategoryCard.js
```diff
@@ -22,9 +24,11 @@ class CategoryCard extends Component {
 
   render() {
     const { name, image, shape } = this.props
+    const cardShape = shape + "Card"
```
Klynger: You should use string interpolation here

react/components/CategoriesHighlights/categoriesHighlights.css
```diff
@@ -0,0 +1,88 @@
...
+  :global(.vtex-carousel) + .squared {
+    top: -50px;
+  }
+
+  :global(.vtex-carousel) + .rectangular {
+    top: 0px;
+  }
+
+  .squaredCard,
+  .squaredCard img {
```
Klynger: I not a good idea write css rules to html elements (applying css rules directly to html elements is slower than applying to a class), instead of doing that you could put a class in the `img` element and then write something like `.squaredCard myImageClass`

react/components/CategoriesHighlights/categoriesHighlights.css
```diff
@@ -0,0 +1,88 @@
...
+
+.squaredCard img {
+  width: 300px;
+}
+
+.rectangularCard img {
+  width: 600px;
+}
+
+@media only screen and (min-width: 320px) and (max-width: 800px) {
```
Klynger: You should use meadia queries with `em` unity instead of `px`

luciannojunior:
@paul0vinicius, you should try to achieve the same results using **only Tachyons**, ok? Directly writing CSS is **extremely discouraged**. 

Also, you should avoid using `px` and use `rem` instead. Check [tachyons.css](https://github.com/vtex/vtex-tachyons/blob/master/tachyons.css)'s dimensions classes.

Happy birthday 🎉

react/components/CategoriesHighlights/components/CategoryCard.js
```diff
@@ -22,20 +24,22 @@ class CategoryCard extends Component {
 
   render() {
     const { name, image, shape } = this.props
+
     return (
       <div
-        className={`vtex-categories-highlighted__category-${shape}-card shadow-1`}>
+        className={`${categoriesHighlights[shape]}Card shadow-1`}>
```
lucasecdb: this won't work, you need to apply the `Card` part of the class to the css modules mapping variable. like so
```suggestion
        className={`${categoriesHighlights[`${shape}Card`]}shadow-1`}>
```
you should also do something similar to the rest of the classes

react/components/CategoriesHighlights/components/CategoryCard.js
```diff
@@ -22,20 +24,22 @@ class CategoryCard extends Component {
 
   render() {
     const { name, image, shape } = this.props
+
     return (
       <div
-        className={`vtex-categories-highlighted__category-${shape}-card shadow-1`}>
+        className={`${categoriesHighlights[shape]}Card shadow-1 ma1`}>
```
lucasecdb: missed here

paul0vinicius: my bad

PR #282, opened 2019-02-06
Now the page won't scroll after changing SKU'S

luciannojunior:
Fixes, in a way, the navigate function on `SKUSelector` regarding scrolling the page after a change in the SKU.

_E de quebra_, also pass the `replace` option as true when the changes of SKU happens in a secondary variation, so when the user clicks on the back button of the browser, it won't keep going back to each previously selected variation. But this behaviour will keep happening for the main variation. All of this was specified by @gabrielgalc 
 
**THIS LAST FEATURE DEPENDS ON [THIS](https://github.com/vtex-apps/render-runtime/pull/218) PR ON RENDER-RUNTIME** .

_I've also done some prettier here and there_
Scrolling the page after the user selected a variation on the Product's page maah 
You can see this behavior on [master](https://storecomponents.myvtex.com). Just select the **Basic Shirt.**

Using [lucis](https://lucis--storecomponents.myvtex.com) and also selecting the **Basic Shirt**.

Naah.

#### Types of changes
- [X] New feature (a non-breaking change which adds functionality)

### DIFF
--- a/react/components/SKUSelector/components/SKUSelector.js
+++ b/react/components/SKUSelector/components/SKUSelector.js
@@ -1,4 +1,5 @@
 import React, { Component } from 'react'
+import { curry } from 'ramda'
 
 import Variation from './Variation'
 import { SKUSelectorPropTypes } from '../utils/proptypes'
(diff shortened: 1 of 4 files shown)


react/components/SKUSelector/components/SKUSelector.js
```diff
@@ -10,18 +11,20 @@ export default class SKUSelector extends Component {
   render() {
     const { selectedId, onSelectSku, mainVariation, secondaryVariation, maxSkuPrice } = this.props
 
+    const onSkuSelected = curry(onSelectSku)
```
ArthurSampaio: Just a doubt, what `curry` does? The ramda doc isnt enough to understand it. :)

lucasecdb: it makes you able to _partially_ apply a function, meaning you can call it with some arguments, and it returns another function with those arguments set, e.g., if you have a function that receives 4 arguments and you call it with 2, it returns another function that only need the other two to be actually executed.

luciannojunior: All of Ramda's functions are _curried_ be default, and it is actually a very nice feature. It kind of "prepare" a function that has more than one argument. It is one of the main reason that Ramda functions' have their parameters kind of "reversed" comparing with our imperative programming view.

If I have `const sum = (a, b) => a + b`, I can get a function `sumWith3` with `const sumWith3 = curry(sum)(3)`. It **prepares** a function for a future apply.

If, for example, I need to map over a collection and sum all elements with 3, I would use my brand new function.

Some functional languages like F# have curry built-in in their core, and every function you create is already curried.

[Read more here](https://stackoverflow.com/questions/36314/what-is-currying)

ArthurSampaio: Nice! Very thanks guys, indeed  is a powerful tool .

PR #283, opened 2019-02-07
Use the minicart's optimistic strategy to add item

estacioneto:
**Related to vtex-apps/minicart#93**

Now, the buy button uses the new minicart's optimistic strategy, explained at the minicart's PR.

The store will feel faster.

[Access the workspace](https://offline--storecomponents.myvtex.com)

#### Types of changes

### DIFF
--- a/react/components/BuyButton/index.js
+++ b/react/components/BuyButton/index.js
@@ -1,6 +1,8 @@
 import find from 'lodash/find'
 import PropTypes from 'prop-types'
 import React, { Component, Fragment } from 'react'
+import { graphql } from 'react-apollo'
+import gql from 'graphql-tag'
 import { injectIntl, intlShape, FormattedMessage } from 'react-intl'
 import ContentLoader from 'react-content-loader'
 import { compose, pick } from 'ramda'
@@ -27,13 +29,13 @@ export class BuyButton extends Component {
   static defaultProps = {
     isOneClickBuy: false,
     available: true,
-  };
+  }
 
   state = {
     isLoading: false,
     isAddingToCart: false,
     timeOut: null,
-  };
+  }
 
   translateMessage = id => this.props.intl.formatMessage({ id: id })
 
@@ -125,10 +127,8 @@ BuyButton.propTypes = {
       /** Quantity of the product sku to be added to the cart */
       quantity: PropTypes.number.isRequired,
       /** Which seller is being referenced by the button */
-      seller: PropTypes.oneOfType([
-        PropTypes.string,
-        PropTypes.number,
-      ]).isRequired,
+      seller: PropTypes.oneOfType([PropTypes.string, PropTypes.number])
+        .isRequired,
       name: PropTypes.string.isRequired,
       price: PropTypes.number.isRequired,
       variant: PropTypes.string,
@@ -161,9 +161,34 @@ BuyButton.propTypes = {
   onAddFinish: PropTypes.func,
 }
 
+const withMutation = graphql(
+  gql`
+    type Item {
+      detailUrl: String
+      id: String
+      imageUrl: String
+      listPrice: Float
+      name: String
+      quantity: Int
+      sellingPrice: Float
+      skuName: String
+    }
+
+    mutation addToCart($item: Item) {
+      addToCart(item: $item) @client
+    }
+  `,
+  {
+    props: ({ mutate }) => ({
+      addToCart: item => mutate({ variables: { item } }),
+    }),
+  }
+)
+
 export default compose(
+  withMutation,
   withToast,
   orderFormConsumer,
   injectIntl,
-  Pixel,
+  Pixel
 )(BuyButton)
--- a/react/components/BuyButton/index.js
+++ b/react/components/BuyButton/index.js
@@ -54,6 +54,7 @@ export class BuyButton extends Component {
 
     if (isOneClickBuy) location.assign(CONSTANTS.CHECKOUT_URL)
 
+    let toastMessage = null
     try {
       const minicartItems = skuItems.map(skuItem => {
         const {
@@ -93,15 +94,16 @@ export class BuyButton extends Component {
         find(linkStateItems, { id: skuItem.skuId })
       )
 
-      // TODO: Minicart should mutate the orderForm and push to Pixel
-      this.toastMessage(success.length >= 1)
+      toastMessage = () => this.toastMessage(success.length >= 1)
     } catch (err) {
       console.error(err)
-      this.toastMessage(false)
+      toastMessage = () => this.toastMessage(false)
     }
 
-    this.setState({ isAddingToCart: false })
-    onAddFinish && onAddFinish()
+    setTimeout(
+      () => this.setState({ isAddingToCart: false }, toastMessage),
+      5e2
+    )
   }
 
   render() {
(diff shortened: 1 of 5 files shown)


react/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "vtex.store-components",
   "scripts": {
-    "test": "vtex-scripts test"
+    "test": "echo FIXME"
```
brunojdo: 🤷‍♂️

react/components/BuyButton/index.js
```diff
@@ -1,15 +1,12 @@
 import find from 'lodash/find'
```
brunojdo: ⛏ ramda? :D

estacioneto: What about do that when migrating to typescript?

react/components/BuyButton/index.js
```diff
@@ -45,44 +42,60 @@ export class BuyButton extends Component {
...
-
-        orderFormContext.refetch()
-        this.toastMessage(success.length >= 1)
-      },
-      () => {
-        this.toastMessage(false)
-      }
+    setTimeout(
+      () => this.setState({ isAddingToCart: false }, toastMessage),
+      5e2
```
luciannojunior: Some reason for that?

react/components/BuyButton/index.js
```diff
@@ -45,44 +42,60 @@ export class BuyButton extends Component {
...
-    const { skuItems, isOneClickBuy, orderFormContext, push } = this.props
+    const { addToCart, skuItems, isOneClickBuy } = this.props
     this.setState({ isAddingToCart: true })
 
-    const variables = {
-      items: skuItems.map(skuItem => {
-        const { skuId } = skuItem
+    if (isOneClickBuy) location.assign(CONSTANTS.CHECKOUT_URL)
+
+    let toastMessage = null
```
luciannojunior:
```suggestion
    let updateToastMessage;
```

react/components/BuyButton/index.js
```diff
@@ -45,44 +42,60 @@ export class BuyButton extends Component {
...
 
   handleAddToCart = async () => {
-    const { skuItems, isOneClickBuy, orderFormContext, push } = this.props
+    const { addToCart, skuItems, isOneClickBuy } = this.props
     this.setState({ isAddingToCart: true })
 
-    const variables = {
-      items: skuItems.map(skuItem => {
-        const { skuId } = skuItem
+    if (isOneClickBuy) location.assign(CONSTANTS.CHECKOUT_URL)
```
luciannojunior: Why before the action? And if something went wrong?

react/components/BuyButton/index.js
```diff
@@ -45,44 +42,60 @@ export class BuyButton extends Component {
...
         return {
-          id: parseInt(skuId),
+          detailUrl,
+          id,
+          imageUrl,
+          listPrice,
+          name,
+          quantity,
+          sellingPrice,
+          skuName,
```
luciannojunior: Why did you lose the `pick`? It got more verbose

react/components/BuyButton/index.js
luciannojunior: The 5e2 in this case

react/components/BuyButton/index.js
```diff
@@ -45,44 +42,60 @@ export class BuyButton extends Component {
   }
 
   handleAddToCart = async () => {
```
luciannojunior: My overall opinion about that is that it is cool, but perharps the handleAddToCart could be partitioned in some other functions... It got bigger

react/components/BuyButton/index.js
estacioneto: @brunojdo didn't like the UX because you don't see the buy-button spinner. So, I added a timeout just to show that.

react/components/BuyButton/index.js
estacioneto: I don't know how it was working before. And back then, it was before the action too...

react/components/BuyButton/index.js
estacioneto: ramda over JS... Someday I'll understand that

react/components/BuyButton/index.js
estacioneto: But it doesn't really `updates` the message. It shows the message 🤔

react/package.json
estacioneto: RIP `vtex-scripts`

react/components/BuyButton/index.js
estacioneto: Forget it. If @luciannojunior is so in love with it, I'll do it 😍

react/components/BuyButton/index.js
luciannojunior: Maybe `showToastMessage`. If some function will perform a side-effect, it is expected that its name indicate that. Anyone would guess that `toastMessage` would hold a string value later, seeing this line, don't you think?

react/components/BuyButton/index.js
luciannojunior: It's not about that, any helper library would do in that case, really.

You turned a transformation of a data structure to return into 1) destructuring the initial DS 2) creating another DS with those destructured props. 

Despite way less LOC's, it's more error prone because you are duplicating the name writing of a bunch of names. Without typechecking or unit tests, it becomes easy to future updates forget something and, if not, it is clearly takes more time

react/components/BuyButton/index.js
luciannojunior: Why not just 500? I'm sure that the majority of JS developers wouldn't know at first how many milliseconds `5e2` represents

react/components/BuyButton/index.js
luciannojunior: Are there more properties being passed than those one listes in `propTypes`? If so, [omit](https://ramdajs.com/docs/#omit) would be even less verbose.

estacioneto: I partially agree, because about this type check, we'll do it in this project. And it won't cover the `pick` function because they are strings. So, if anyone changes some string value, the type check will do nothing. Anyway, I've used the `pick` function, as you asked, senpai :)

react/components/BuyButton/index.js
estacioneto: That's true. I'll change it (and I promise I won't use this argument against ramda 🙊)

react/components/BuyButton/index.js
estacioneto: Done

react/components/BuyButton/index.js
estacioneto: About, omit... Maybe there are more properties. But even if there aren't, I want to make sure only these selected properties go to the minicart. I don't wanna risk someone adding stuff and breaking the link-state.

luciannojunior: Ok. 

About the typechecking of `pick`, that is not actually true, as you can see [here](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/ramda/index.d.ts#L1751) you can (and our platforms will enforce you of that) pass the type parameters expected for `pick`

estacioneto: Oh cool! I just feel a little bit curious about this `If the key does not exist, the property is ignored`. Because if I mistyped a property, It will warn me or just ignore it? I'll play with it someday ;)

PR #284, opened 2019-02-07
add BuyButton on press callbacks (start & finish)

jgfidelis:
I'd like to in my components, know when the buybutton has been pressed and has finished processing. To make adjustments on the parent component while the mutation is called.

Add props `onAddPress` and `onAddFinish` that are called at the start and end of the handle button click method

#### Types of changes

### DIFF
--- a/react/components/BuyButton/index.js
+++ b/react/components/BuyButton/index.js
@@ -45,9 +45,9 @@ export class BuyButton extends Component {
   }
 
   handleAddToCart = async () => {
-    const { skuItems, isOneClickBuy, orderFormContext, push, onAddPress, onAddFinish } = this.props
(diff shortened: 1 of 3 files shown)


react/components/BuyButton/index.js
```diff
@@ -45,8 +45,9 @@ export class BuyButton extends Component {
   }
 
   handleAddToCart = async () => {
-    const { skuItems, isOneClickBuy, orderFormContext, push } = this.props
+    const { skuItems, isOneClickBuy, orderFormContext, push, onAddPress, onAddFinish } = this.props
```
lbebber: perhaps switch from `onAddPress` to `onAddStart`, in order to be in line with `onAddFinish`

salesfelipe: up

PR #316, opened 2019-02-18
Fix product images refresh

lucasecdb:
Fix precondition on state reset of the `Carousel` component.

Due to the wrong precondition check, the component was "flashing", even though the images and thumbnails were already loaded.

[Workspace](https://lucas--instoreqa.myvtex.com/instore/1177?skuId=3333). Or link this branch in any workspace with `vtex link`.

#### Screenshots

Before:
before

After:
after

#### Types of changes

### DIFF
--- a/react/components/ProductImages/components/Carousel/index.js
+++ b/react/components/ProductImages/components/Carousel/index.js
@@ -68,6 +68,8 @@ class Carousel extends Component {
 
   componentWillUnmount() {
     window.removeEventListener('resize', this.debouncedRebuildOnUpdate)
+
+    this.debouncedRebuildOnUpdate.flush()
   }
 
   onSlideChange = () => {
--- a/react/components/ProductImages/components/Carousel/index.js
+++ b/react/components/ProductImages/components/Carousel/index.js
@@ -24,8 +24,6 @@ const initialState = {
 }
 
 class Carousel extends Component {
-  _mounted = false
-
   state = initialState
 
   thumbSwiper = React.createRef()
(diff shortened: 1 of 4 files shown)


react/components/ProductImages/components/Carousel/index.js
```diff
@@ -281,7 +262,7 @@ class Carousel extends Component {
             }
           )}
         >
-          <Swiper {...galleryParams}>
+          <Swiper {...galleryParams} ref={this.gallerySwiper}>
```
felippenardi: Why is the `ref`s needed now?

lucasecdb: to remove the need for this code
```javascript
linkGallery = () => {
  const _this = this
  return function() {
    _this.setState({ gallerySwiper: this })
  }
}
```

This also removes the need to keep the `swiper` instance in the component state.

react/components/ProductImages/components/Carousel/index.js
```diff
@@ -144,10 +140,11 @@ class Carousel extends Component {
...
 
   componentDidUpdate(prevProps) {
-    const { gallerySwiper, loaded, activeIndex } = this.state
-    const { isVideo } = this
+    const { loaded, activeIndex } = this.state
+    const isVideo = this.isVideo
+    const gallerySwiper = path(['swiper'], this.gallerySwiper.current)
 
-    if (prevProps.slides !== this.props.slides) {
+    if (!equals(prevProps.slides, this.props.slides)) {
```
felippenardi: What condition does `!equals` verifies that `!==` is not enough?

lucasecdb: `!==` evaluates to true if the elements have different *references*, while `!equals` compares the values in the objects inside the array.

PR #320, opened 2019-02-19
Add Snapshot test to ProductPrice

JuanAlmeida12:
As the title says.

#### Types of changes


react/__tests__/components/ProductPrice.test.js
```diff
@@ -0,0 +1,37 @@
+import React from 'react'
+import ProductPrice from './../../ProductPrice'
+import { render } from 'enzyme'
+
+describe("<ProductPrice />", () => {
+  let renderComponent = null
```
GustavoMA: Instead of using a global variable, why not having a function that generates and returns the component to be tested

react/__tests__/components/ProductPrice.test.js
```diff
@@ -0,0 +1,37 @@
+import React from 'react'
+import ProductPrice from './../../ProductPrice'
+import { render } from 'enzyme'
```
klzns: I would rather use `react-testing-library`.

It makes it easier to avoid testing implementation details.

react/__tests__/components/ProductPrice.test.js
```diff
@@ -0,0 +1,37 @@
...
+      return render(<ProductPrice {...props} />, { context })
+    }
+  })
+
+  it('should be mount', () => {
+    expect(renderComponent()).toBeDefined()
+  })
+
+  it('should match snapshot', () => {
+    expect(renderComponent()).toMatchSnapshot()
```
klzns: It's more useful and clear when instead of using snapshot, we expect to find a specific element on the rendered DOM.

react/__tests__/components/ProductPrice.test.js
klzns: I wrote some docs on how to setup:
https://docs.google.com/document/d/1RRjBWTtA9tO_6ci29XUbtPCR5_QkpcGM6q_VRUwstuc/edit?usp=sharing

react/__tests__/components/ProductPrice.test.js
GustavoMA: Why not instead of having a global variable, have a function that generate and returns the component to be tested?

GustavoMA:
Delete the helpers folder. It's making the Travis fail.
Also, add Travis in the README :)

PR #321, opened 2019-02-19
Add snapshot tests for BuyButton

theomoura:
As title says

#### Types of changes


react/__tests__/components/BuyButton.test.js
```diff
@@ -0,0 +1,44 @@
...
+
+    return {
+      component,
+    }
+  }
+
+  beforeEach(() => {})
+
+  it('should be rendered', () => {
+    wrapper = renderComponent()
```
GustavoMA:
```suggestion
    const wrapper = renderComponent()
```
Let's avoid possible colateral effects of using a global variable. The same for the other `it`s

react/__tests__/components/BuyButton.test.js
```diff
@@ -0,0 +1,44 @@
...
+describe('<BuyButton />', () => {
+  let wrapper
+
+  const renderComponent = customProps => {
+    const props = {
+      ...customProps,
+      showToast: () => {},
+    }
+
+    const component = shallow(<BuyButton {...props}> Test </BuyButton>)
```
GustavoMA: We need to check out why the `render`  function is returning a strange Snapshop.

react/__mocks__/vtex.styleguide.js
```diff
@@ -0,0 +1,9 @@
+import React, { Fragment } from 'react'
+
+export function Button(props) {
+  return <div> {props.children} </div>
```
GustavoMA:
```suggestion
  return <div class="button"> {props.children} </div>
```
Or
```suggestion
  return <button type="button"> {props.children} </button>
```
Or mocks should be a little more inteligent to create better snapshots.

react/__mocks__/react-intl.js
```diff
@@ -0,0 +1,13 @@
+import React, { Fragment } from 'react'
```
GustavoMA: Probably ou can avoid mocking `react-intl` using something like this https://github.com/vtex-apps/shelf/blob/master/react/__test__/Shelf.test.js#L136

react/package.json
```diff
@@ -43,11 +43,12 @@
...
     "babel-preset-react-app": "^7.0.1",
-    "enzyme": "^3.3.0",
-    "enzyme-adapter-react-16": "^1.1.1",
+    "enzyme": "^3.4.1",
+    "enzyme-adapter-react-16": "^1.2.0",
     "enzyme-to-json": "^3.3.5",
     "jest": "^24.1.0",
-    "jest-transform-graphql": "^2.1.0"
+    "jest-transform-graphql": "^2.1.0",
+    "identity-obj-proxy": "^3.0.0"
```
luciannojunior: Needed?

react/__tests__/components/BuyButton.test.js
theomoura: Its was not returning strange snapshot, since the component was only an button, it was returning what was in my mock. Since I want to make a snapshot of a simple component, the more appropiate is to use `shallow`

react/package.json
theomoura: Yeah, this mocks the css modules for my tests

PR #323, opened 2019-02-19
Snapshot tests for animation component

theomoura:
As title says.

#### Types of changes

### DIFF
--- a/react/__tests__/components/Animation.test.js
+++ b/react/__tests__/components/Animation.test.js
@@ -1,27 +1,37 @@
 import React from 'react'
 
 import Animation from '../../Animation'
-import { renderWithIntl } from '../../__helpers__/intl-helper'
+import { render } from 'enzyme'
 
 describe('<Animation /> component', () => {
-  let wrapper
-
   const renderComponent = customProps => {
     const comp = <Animation {...customProps}> Test </Animation>
 
-    return renderWithIntl(comp)
+    return render(comp)
   }
 
-  beforeEach(() => {
-    wrapper = renderComponent()
-  })
-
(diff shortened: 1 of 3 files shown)


react/__tests__/components/Animation.test.js
```diff
@@ -0,0 +1,27 @@
...
+  beforeEach(() => {
+    wrapper = renderComponent()
+  })
+
+  it('should be rendered', () => {
+    expect(wrapper).toBeDefined()
+    expect(wrapper).toBeTruthy()
+  })
+
+  it('should match snapshot', () => {
```
GustavoMA: Shouldn't be a test for each animation type? Since each one has different styles.

react/__tests__/components/Animation.test.js
```diff
@@ -0,0 +1,27 @@
+import React from 'react'
+
+import Animation from '../../Animation'
+import { renderWithIntl } from '../../__helpers__/intl-helper'
+
+describe('<Animation /> component', () => {
+  let wrapper
```
GustavoMA: I would avoid using `let` and reassigning this variable.

react/__tests__/components/Animation.test.js
```diff
@@ -0,0 +1,27 @@
...
+
+    return renderWithIntl(comp)
+  }
+
+  beforeEach(() => {
+    wrapper = renderComponent()
+  })
+
+  it('should be rendered', () => {
+    expect(wrapper).toBeDefined()
```
GustavoMA: `toBeTruthy` already tests if it isn't undefined. So, no need to use `toBeDefined`.

PR #326, opened 2019-02-19
Snapshots tests for Container component

theomoura:
As title says.

#### Types of changes

### DIFF
--- a/react/__tests__/components/Container.test.js
+++ b/react/__tests__/components/Container.test.js
@@ -1,0 +1,22 @@
+import React from 'react'
+
+import Container from '../../Container'
+import { render } from 'enzyme'
+
+describe('<Container /> component', () => {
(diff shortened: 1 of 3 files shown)


react/__tests__/components/Container.test.js
```diff
@@ -0,0 +1,22 @@
...
+import { render } from 'enzyme'
+
+describe('<Container /> component', () => {
+  const renderComponent = customProps => {
+    const comp = <Container {...customProps}> Test </Container>
+
+    return render(comp)
+  }
+
+  const wrapper = renderComponent()
```
lbebber: I think it'd be better to run this function and create this wrapper on each test instead

theomoura: nice tip

PR #328, opened 2019-02-20
Add GradientCollapse snapshot tests

GustavoMA:

As described in the title.

#### Types of changes

### DIFF
--- a/react/__tests__/components/GradientCollapse.test.js
+++ b/react/__tests__/components/GradientCollapse.test.js
@@ -1,0 +1,26 @@
+import React from 'react'
+
+import { renderWithIntl } from '../../__helpers__/intl-helper'
+import GradientCollapse from '../../GradientCollapse'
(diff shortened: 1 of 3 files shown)


react/__tests__/components/GradientCollapse.test.js
```diff
@@ -0,0 +1,26 @@
+import React from 'react'
+
+import { renderWithIntl } from '../../__helpers__/intl-helper'
```
lucasecdb: this is just a suggestion, as this is probably in several tests now, but what do you think about setting the jest config to use this `__helpers__` folder as a root for modules? That way we could import them as
```js
import { renderWithIntl } from 'intl-helper'
```
and probably rename the folder to something like `testUtils`

GustavoMA: Done in #333

PR #333, opened 2019-02-20
Better use of helpers in tests

GustavoMA:
#### Types of changes


react/__tests__/components/DiscountBadge.test.js
```diff
@@ -16,17 +16,17 @@ describe('<DiscountBadge />', () => {
   }
 
   it('should be able to mount and not break', () => {
-    const comp = renderComponent() 
+    const comp = renderComponent()
```
lbebber: ⛏ this abbreviation is a bit unnecessary, why not just name it `component`?

react/package.json
```diff
@@ -57,6 +57,10 @@
   },
   "jest": {
     "verbose": true,
+    "moduleDirectories": [
+      "node_modules",
+      "__helpers__"
```
lucasecdb: add `<rootDir>` as a prefix

react/package.json
```diff
@@ -57,6 +57,10 @@
   },
   "jest": {
     "verbose": true,
+    "modulePaths": [
+      "<rootDir>/node_modules",
```
lucasecdb: you dont need the `node_modules` here

PR #345, opened 2019-02-25
Feature/user address

lbebber:
Adds `UserAddress` component, intended for inclusion on the page's header.

screen shot 2019-02-25 at 18 00 58

screen shot 2019-02-25 at 18 01 10

### DIFF
--- a/react/components/UserAddress/ChangeAddressIcon.js
+++ b/react/components/UserAddress/ChangeAddressIcon.js
@@ -1,0 +1,13 @@
+import React from 'react'
+
+const ChangeAddressIcon = () => {
+  return (
+    <svg width="21" height="27" viewBox="0 0 21 27" fill="none" xmlns="http://www.w3.org/2000/svg">
+      <path d="M19.9655 10.4828C19.9655 16.3448 10.4828 26 10.4828 26C10.4828 26 1 16.431 1 10.4828C1 4.44828 5.91379 1 10.4828 1C15.0517 1 19.9655 4.53448 19.9655 10.4828Z" stroke="currentColor" stroke-width="2" stroke-miterlimit="10" stroke-linecap="round" stroke-linejoin="round"/>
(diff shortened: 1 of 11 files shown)


react/components/UserAddress/ChangeAddressIcon.js
```diff
@@ -0,0 +1,13 @@
+import React from 'react'
+
+const ChangeAddressIcon = () => {
+  return (
+    <svg width="21" height="27" viewBox="0 0 21 27" fill="none" xmlns="http://www.w3.org/2000/svg">
+      <path d="M19.9655 10.4828C19.9655 16.3448 10.4828 26 10.4828 26C10.4828 26 1 16.431 1 10.4828C1 4.44828 5.91379 1 10.4828 1C15.0517 1 19.9655 4.53448 19.9655 10.4828Z" stroke="currentColor" stroke-width="2" stroke-miterlimit="10" stroke-linecap="round" stroke-linejoin="round"/>
```
lucasecdb: shouldn't `stroke-width`, `stroke-miterlimit`, `stroke-linecap` and `stroke-linejoin` be capitalized like `strokeWidth`?

react/components/UserAddress/ChangeAddressIcon.js
```diff
@@ -0,0 +1,13 @@
...
+      <path d="M10.4827 13.069C11.911 13.069 13.0689 11.9111 13.0689 10.4828C13.0689 9.05443 11.911 7.89655 10.4827 7.89655C9.05437 7.89655 7.89648 9.05443 7.89648 10.4828C7.89648 11.9111 9.05437 13.069 10.4827 13.069Z" stroke="currentColor" stroke-width="2" stroke-miterlimit="10" stroke-linecap="round" stroke-linejoin="round"/>
```
lucasecdb: same as above

react/components/UserAddress/index.js
```diff
@@ -0,0 +1,111 @@
+import React  from 'react'
+import PropTypes from 'prop-types'
+import { orderFormConsumer, contextPropTypes } from 'vtex.store-resources/OrderFormContext'
+import ChangeAddressIcon from './ChangeAddressIcon'
+
+import Container from '../Container'
+
+import { FormattedMessage } from 'react-intl'
```
lucasecdb: place this import before the relative ones

react/components/UserAddress/ChangeAddressIcon.js
lbebber: It works like this, so 🤷‍♂️

lucasecdb: I think react warns if the props are like that, but I guess it's fine

PR #352, opened 2019-03-07
Update store-components to use react-testing-library

JuanAlmeida12:
As the title says.

#### Types of changes


react/package.json
```diff
@@ -39,39 +39,37 @@
...
+    "react-testing-library": "^6.0.0",
+    "regenerator-runtime": "^0.13.1"
   },
   "vtexScriptsOverride": {
     "srcPath": "."
   },
   "jest": {
     "verbose": true,
-    "modulePaths": [
+    "moduleDirectories": [
```
GustavoMA: It's better to keep using `modulePaths`.  `moduleDirectories` tries to find the folders recursively, so if there is two folders named `testUtils` depending how the search is done the wrong folder may be used.

react/package.json
```diff
@@ -39,39 +39,37 @@
...
+    "react-testing-library": "^6.0.0",
+    "regenerator-runtime": "^0.13.1"
   },
   "vtexScriptsOverride": {
     "srcPath": "."
   },
   "jest": {
     "verbose": true,
     "modulePaths": [
+      "node_modules",
```
GustavoMA: No need to add `node_modules` using `modulePaths` :)

PR #356, opened 2019-03-12
 Add behavior tests to buybutton and fix snapshots.

JuanAlmeida12:
As the title says.

#### Types of changes


react/__tests__/components/BuyButton.test.js
```diff
@@ -9,11 +10,19 @@ describe('<BuyButton />', () => {
       ...customProps,
     }
 
-    const comp = <BuyButton {...props}> Test </BuyButton>
+    const comp = <BuyButton {...props}>Test</BuyButton>
 
     return render(comp)
   }
 
+  function wait(ms) {
```
GustavoMA: Looks like you are not using this `wait` function.

react/__tests__/components/BuyButton.test.js
```diff
@@ -34,4 +43,32 @@ describe('<BuyButton />', () => {
     const { asFragment } = renderComponent({ available: false })
     expect(asFragment()).toMatchSnapshot()
   })
+
+  it('should call onAddStart', () => {
```
GustavoMA: I think both test cases can be merged into one that will garantee that `onAddStart` is called then `onAddEnd` is called.

react/__mocks__/vtex.store-resources/OrderFormContext.js
```diff
@@ -1,5 +1,85 @@
 import React from 'react'
+import PropTypes from 'prop-types'
+import OrderForm from 'orderForm'
 
-export function orderFormConsumer(Comp) {
-  return Comp
+export const contextPropTypes = PropTypes.shape({
```
jgfidelis: we should make a `@types/vtex-io` and export this stuff there, then, import it here...
my 2 cents

react/__tests__/components/BuyButton.test.js
```diff
@@ -34,4 +35,32 @@ describe('<BuyButton />', () => {
...
+    const onAddStart = jest.fn()
+    const { getByText } = renderComponent({
+      available: true,
+      skuItems: [],
+      onAddStart,
+    })
+
+    fireEvent.click(getByText(/test/i))
+    expect(onAddStart).toHaveBeenCalled()
```
theomoura: You should check if this is called once.

react/__tests__/components/BuyButton.test.js
```diff
@@ -34,4 +35,28 @@ describe('<BuyButton />', () => {
...
+        available: true,
+        skuItems: [],
+        onAddStart,
+        onAddFinish,
+      },
+      buttonText
+    )
+
+    fireEvent.click(getByText(buttonText))
+    setTimeout(() => {
```
klzns: Why are you using setTimeout?

JuanAlmeida12: Because this `onClick` is asynchronous and don't modify the DOM so I can't use `wait` or `waitForElement` if you have any suggestion, I will accept.

PR #361, opened 2019-03-13
Set focus on search bar when search icon is clicked [mobile]

theomoura:
Set the focus of the search bar when user clicks on the search icon on mobile devices.

The user had to click twice on the search to actually search something.

[Access](https://barfocus--storecomponents.myvtex.com), set to mobile devices and refresh page

#### Types of changes

### DIFF
--- a/react/__mocks__/vtex.render-runtime.js
+++ b/react/__mocks__/vtex.render-runtime.js
@@ -1,8 +1,8 @@
 import React from 'react'
 
-export function withRuntimeContext(Comp) {
-  return Comp
-}
+export const withRuntimeContext = Comp => props => (
+  <Comp {...props} runtime={{ hints: { mobile: false } }} />
+)
 
 export const Link = ({ children }) => <a>{children}</a>
 
--- a/react/__tests__/components/__snapshots__/SearchBar.test.js.snap
+++ b/react/__tests__/components/__snapshots__/SearchBar.test.js.snap
@@ -24,7 +24,6 @@ exports[`<SearchBar /> should match snapshot 1`] = `
               id="downshift-1-input"
               placeholder="Search"
               role="combobox"
-              runtime="[object Object]"
               suffix="[object Object]"
               value=""
             />
--- a/react/__tests__/components/__snapshots__/SearchBar.test.js.snap
+++ b/react/__tests__/components/__snapshots__/SearchBar.test.js.snap
@@ -24,6 +24,7 @@ exports[`<SearchBar /> should match snapshot 1`] = `
               id="downshift-1-input"
               placeholder="Search"
               role="combobox"
+              runtime="[object Object]"
               suffix="[object Object]"
               value=""
             />
--- a/react/components/SearchBar/components/AutocompleteInput.js
+++ b/react/components/SearchBar/components/AutocompleteInput.js
@@ -1,75 +1,83 @@
-import React, { useLayoutEffect, useRef } from 'react'
+import React, { Component } from 'react'
 import PropTypes from 'prop-types'
 import classNames from 'classnames'
 
 import { Input } from 'vtex.styleguide'
 import { IconClose, IconSearch } from 'vtex.store-icons'
-import { useRuntime } from 'vtex.render-runtime'
+import { withRuntimeContext } from 'vtex.render-runtime'
 
 import styles from '../styles.css'
 
 /** Midleware component to adapt the styleguide/Input to be used by the Downshift*/
-const AutocompleteInput = (
-  onGoToSearchPage,
-  onClearInput,
-  compactMode,
-  value,
-  hasIconLeft,
-  iconClasses,
-  ...restProps
-) => {
-  let inputRef = useRef(null)
-  const { mobile: isMobile } = useRuntime().hints
+class AutocompleteInput extends Component {
+  constructor(props) {
+    super(props)
+    this.inputRef = React.createRef()
+  }
 
-  const changeClassInput = () => {
+  changeClassInput = () => {
+    const { compactMode } = this.props
     if (compactMode) {
-      inputRef.current.placeholder = ''
-      inputRef.current.classList.add(styles.paddingInput)
+      this.inputRef.current.placeholder = ''
+      this.inputRef.current.classList.add(styles.paddingInput)
     }
   }
 
-  useLayoutEffect(() => {
-    changeClassInput()
-    isMobile && inputRef.current.focus()
-  }, [])
+  componentDidMount() {
+    const { runtime } = this.props
+    this.changeClassInput()
+    runtime.hints.mobile && this.inputRef.current.focus()
+  }
 
-  const suffix = (
-    <span
-      className={`${iconClasses} flex items-center pointer`}
-      onClick={() => value && onClearInput()}
-    >
-      {value ? (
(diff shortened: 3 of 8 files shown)


react/__tests__/components/__snapshots__/SearchBar.test.js.snap
```diff
@@ -24,6 +24,7 @@ exports[`<SearchBar /> should match snapshot 1`] = `
               id="downshift-1-input"
               placeholder="Search"
               role="combobox"
+              runtime="[object Object]"
```
felippenardi: What is this snapshot useful for?

react/components/SearchBar/components/AutocompleteInput.js
```diff
@@ -23,7 +24,9 @@ export default class AutocompleteInput extends Component {
   }
 
   componentDidMount() {
+    const { runtime } = this.props
     this.changeClassInput()
+    runtime.hints.mobile && this.inputRef.current.focus()
```
felippenardi: Is it worth checking if `runtime.hints.mobile` is not undefined to avoid runtime errors?

theomoura:
> The focus event is bind to the component did mount, not a click in the mobile.
> 
> Is it safe to assume that whenever this component is mounted it will be because the user clicked on the search icon?
> 
> Otherwise, this focus may happen in unexpected situations.

The focus event is set to only focus when the component is mounted on mobile devices, and this is what we want since it is mounted when the user clicks on the search icon. On other devices this behaviour will not occur.

react/__tests__/components/__snapshots__/SearchBar.test.js.snap
theomoura: It tests if there was any structure change to the component.

react/components/SearchBar/components/AutocompleteInput.js
theomoura: Good idea!

react/__tests__/components/__snapshots__/SearchBar.test.js.snap
felippenardi: Testing if it contains `object Object` doesn't seem like a useful test for me—quite the opposite, it bricks the evolution of the component with an assertion that does not sums up any confidence to whether the component is working as expected or not.

theomoura: I uderstand, but since the snapshot is automatically generated and my mocks only states that runtime is a HOC that returns a component i don't know what else i can do to improve it. If you have any suggestion i will be glad to hear.

react/components/SearchBar/components/AutocompleteInput.js
```diff
@@ -23,7 +25,10 @@ export default class AutocompleteInput extends Component {
   }
 
   componentDidMount() {
+    const { runtime } = this.props
     this.changeClassInput()
+    console.log('device', path(['hints', 'mobile'], runtime))
```
matheusps: Remove the `console.log`

react/components/SearchBar/components/AutocompleteInput.js
```diff
@@ -1,14 +1,16 @@
...
 
 import { Input } from 'vtex.styleguide'
 import { IconClose, IconSearch } from 'vtex.store-icons'
+import { withRuntimeContext } from 'vtex.render-runtime'
 
 import styles from '../styles.css'
 
 /** Midleware component to adapt the styleguide/Input to be used by the Downshift*/
-export default class AutocompleteInput extends Component {
+class AutocompleteInput extends Component {
```
matheusps: I would definitely recommend changing it to a functional component and use the hook `useRuntime`. Usage of HOCs makes the component tree look bulkier.
Buuuut, i don't think that this component should care about mobile or desktop devices. Should be a concern of the components that make use of it.

react/components/SearchBar/components/AutocompleteInput.js
```diff
@@ -23,7 +25,10 @@ export default class AutocompleteInput extends Component {
   }
 
   componentDidMount() {
+    const { runtime } = this.props
     this.changeClassInput()
+    console.log('device', path(['hints', 'mobile'], runtime))
+    path(['hints', 'mobile'], runtime) && this.inputRef.current.focus()
```
matheusps: I do not think that this approach looks right. Bind this behavior every time on mobile mode, could compromise the component reusability in different scenarios and become a future issue. This should be configured via props, example: `<Search autoFocus={true} />`

react/__mocks__/vtex.render-runtime.js
```diff
@@ -1,8 +1,8 @@
 import React from 'react'
 
-export function withRuntimeContext(Comp) {
-  return Comp
-}
+export const withRuntimeContext = Comp => props => (
```
matheusps: Note: In case of you agreeing with my suggestions, this would not be necessary anymore.

react/components/SearchBar/components/AutocompleteInput.js
theomoura: The focus on mobile should be something that always happens, it would be strange to have a props to decide if should focus or not.

react/components/SearchBar/components/AutocompleteInput.js
theomoura: The `DownsiftComponent` does not work when using a function component, already tried it.

react/__mocks__/vtex.render-runtime.js
theomoura: This mock is been used for another components as well.

react/components/SearchBar/components/AutocompleteInput.js
matheusps: For example... Another component comes up and wants to reuse this search bar. In this new component, the search bar is always there... Think about it. This behavior would be undesirable.

react/components/SearchBar/components/AutocompleteInput.js
matheusps: Strange... But I believe if you say so.

react/__mocks__/vtex.render-runtime.js
matheusps: Oh, okay 😄

react/components/SearchBar/components/AutocompleteInput.js
theomoura: You made my mind, thinking this way it is a good choice to make this a choice by props

react/components/SearchBar/components/SearchBar.js
```diff
@@ -116,5 +117,5 @@ SearchBar.propTypes = {
   /** Identify if the search icon is on left or right position */
   hasIconLeft: PropTypes.bool,
   /** Custom classes for the search icon */
-  iconClasses: PropTypes.string 
-}
\ No newline at end of file
+  iconClasses: PropTypes.string,
```
matheusps: Shouldn't you include autoFocus propTypes here as well ?🤔

react/components/SearchBar/components/AutocompleteInput.js
```diff
@@ -102,3 +105,5 @@ AutocompleteInput.propTypes = {
   /** Custom classes for the search icon */
```
matheusps: The same

PR #376, opened 2019-03-21
add fallback to add item in store-graphql for now

jgfidelis:
#### Types of changes


react/components/BuyButton/index.js
```diff
@@ -76,26 +77,49 @@ export class BuyButton extends Component {
...
       } = await addToCart(minicartItems)
 
-      const success =
-        linkStateItems &&
-        skuItems.filter(
-          skuItem => !!linkStateItems.find(({ id }) => id === skuItem.skuId)
+      let success = null
+      if (!linkStateItems) {
+        // minicart does not have link state implemented, calling graphql directly
+        // SHOULD BE ERASED ON THE NEAR FUTURE
```
brunojdo: I don't think we need erase in the future. I think this is a good fallback to secure the independent of the components

PR #384, opened 2019-03-28
Fix labels to selling price and list price.

JuanAlmeida12:
As the title says.

Label for list price with ~strikethrough~, and selling price with label `To` even if not have an offer.

Before:
summaryTo
summarylistout

After: 

summaryToFix
summarylistoutfix

#### Types of changes

### DIFF
--- a/react/components/ProductPrice/index.js
+++ b/react/components/ProductPrice/index.js
@@ -74,8 +74,6 @@ class Price extends Component {
 
     const differentPrices = showListPrice && sellingPrice !== listPrice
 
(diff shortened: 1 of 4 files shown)


react/components/ProductPrice/index.js
```diff
@@ -74,6 +74,8 @@ class Price extends Component {
 
     const differentPrices = showListPrice && sellingPrice !== listPrice
 
+    console.log(sellingPrice, listPrice)
```
jgfidelis: console log

PR #385, opened 2019-03-28
Remove Warning and Fixing tests

JuanAlmeida12:

Change `fill-available` to use `stretch`.

Warning in build about use `fill-available`.

[Acess this workspace](https://adminstyleguide--storecomponents.myvtex.coma)

#### Types of changes

### DIFF
--- a/react/components/ProductImages/styles.css
+++ b/react/components/ProductImages/styles.css
@@ -14,7 +14,7 @@
 .loader {
   height: -moz-available;
   height: -webkit-fill-available;
-  height: fill-available;
+  height: stretch;
 }
 
 .imageBlur30 {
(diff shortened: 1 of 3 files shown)


react/components/ProductImages/styles.css
```diff
@@ -14,7 +14,7 @@
 .loader {
   height: -moz-available;
   height: -webkit-fill-available;
-  height: fill-available;
+  height: stretch;
```
theomoura: Did you test it on IE?

gustavorosolem: Why use this unusual height attribute? A think it's better change to something else, like a flexbox.
This attr it's not safe to use.

PR #390, opened 2019-04-01
Add Title Component

sahanljc:
We introduce a title component which can be used to render titles in the store, or in other blocks with center, left and right alignment options.

We need to add a title to several `CategoriesHighlights` sections in the homepage. This `Title` component can be added directly to the store's `blocks.json` file or preferably used in the `CategoriesHighlights` component to render the title.

1. Allow `title` block in the `store.home` block in `vtex.store`.
2. Add `title` block to the `store.home` block of the store.
3. Change configuration values in the CMS > Storefront to observe the effects.

title

#### Types of changes
- [x] Requires change to documentation, which has been updated accordingly.


react/components/Title/index.js
```diff
@@ -0,0 +1,68 @@
+import PropTypes from 'prop-types'
+import React, { Component } from 'react'
+
+import { LEFT, RIGHT, CENTER } from './constants.js'
+import title from './title.css'
+
+/**
+ * Name component. Show name and relevant SKU information of the Product Summary
```
matheusps: This description does not look quite right 🤔

react/components/Title/index.js
```diff
@@ -0,0 +1,68 @@
...
+ */
+class Title extends Component {
```
matheusps: Using a class here makes the component look bulkier for no reason, in my opinion. (A function would look better)

react/components/Title/index.js
```diff
@@ -0,0 +1,68 @@
...
+  static propTypes = {
+    /** Text to be displayed */
+    content: PropTypes.string,
```
matheusps: Why not just render `children` instead of using this prop `content`? 🤔

react/components/Title/index.js
```diff
@@ -0,0 +1,68 @@
...
+    alignment: CENTER,
+  }
+
+  render() {
+    const { content, alignment } = this.props
+
+    return (
+      <div className={`${title.titleContainer} flex justify-between`}>
+        <div className={`${title.titleBox} w-100 ma1 c-muted-1`}>
+          {alignment == LEFT && <h3 className="tl t-heading-3">{content}</h3>}
```
matheusps: Okay... I think this would be cleaner if you just:
```js
  ...
   import classNames from 'classnames'
  ...
  cosnt titleClasses = classNames(styles.title, 't-heading-3', {
    tl: align === LEFT,
    tc: align === CENTER,
    tr: align === RIGHT
  })
  ...
  <h3 className={titleClasses}>{content}</h3>
  ...
```

react/components/Title/index.js
```diff
@@ -0,0 +1,68 @@
+import PropTypes from 'prop-types'
+import React, { Component } from 'react'
+
+import { LEFT, RIGHT, CENTER } from './constants.js'
+import title from './title.css'
```
matheusps: Use `styles` instead of `title`

react/components/Title/index.js
sahanljc: We just want to render a heading (E.g. `<h3>`). If we use `props.children`, how can we make sure that we have only one of this kind of element? We were referring to the `ProductName` component here.

react/components/Title/index.js
sahanljc: Done. :sunglasses:

react/components/Title/index.js
sahanljc: You are correct. Thank you. :+1:

react/components/Title/index.js
sahanljc: Thank you for this! Done. :smile:

react/components/Title/index.js
sahanljc: :+1:

react/components/Title/README.md
```diff
@@ -0,0 +1,42 @@
...
+    "component": "Title"
+  }
+```
+
+### Configuration
+
+Through the Storefront, you can change the `Title`'s behavior and interface. However, you also can make in your theme app, as Dreamstore does.
+
+| Prop name | Type | Description |
+| --- | --- | --- |
```
matheusps: Missing the `level` prop @ docs 🤓

sahanljc: Oops.. Done! :+1:

klzns:
Hi, @sahanljc! You can use the component [`rich-text`](https://github.com/vtex-apps/rich-text) instead. It solves the same problem in a more generic way.

PR #404, opened 2019-04-05
Fix sku not being selected on change

lucasecdb:
Update the sku selector state to match it's prop `skuSelected`, instead of doing that only on mount and updating manually later.

This PR also includes the new linter version and updates all components with prettier, to see the actual changes I've made [click here](https://github.com/vtex-apps/store-components/pull/404/files/c18ee07bf336dc85c423c736018628b54508a0a7..3f9a7a0fdc64381d8e106e1f14b5054285bd155d).

When selecting a different sku on the client, the selection wasn't updated according to the props it received.

[Store Components](https://lucas--storecomponents.myvtex.com/classic-shoes/p?skuId=35) and [PH Mexico](https://lucas--phmex.myvtex.com/super_suprema/p).

#### Types of changes


react/components/SKUSelector/index.js
```diff
@@ -13,107 +14,111 @@ import {
...
+  onSKUSelected,
+}) => {
+  const [mainVariation, setMainVariation] = useState(null)
+  const [secondaryVariation, setSecondaryVariation] = useState(null)
+
+  const { setQuery } = useRuntime()
+
+  const parsedItems = useMemo(() => skuItems.map(parseSku), [skuItems])
+
+  const buildVariations = useCallback(
```
jgfidelis: shouldn't it be `useMemo`?

lucasecdb: no, it's a function

jgfidelis: HMMM vdd viagem minha :D

PR #407, opened 2019-04-09
Add fillRowWithEmptyCards Option

sahanljc:
Allow users to choose to fill category cards up to the row length. This value defaults to true.

We plan to use `CategoriesHighlights` in the `vtex.search-result` app to show the category tree response. The current behaviour of this component is to show empty cards, if the categories do not add up to complete the row. We do not want this to happen in the search result page.

Try the search result page [here](https://sahan--biscoindqa.myvtex.com/south-co/b).

##### without fill
ch-without-fill
##### with fill
ch-with-fill

#### Types of changes
- [x] Requires change to documentation, which has been updated accordingly.


victorhmp:
Closing this due to it being stale.

PR #418, opened 2019-04-12
Fix/container size

lbebber:
Decreases horizontal padding for Container on wide screen sizes.
Adds `container` CSS handle to Container.

#### Types of changes

### DIFF
--- a/react/components/Container/Container.css
+++ b/react/components/Container/Container.css
@@ -1,0 +1,1 @@
+.container {}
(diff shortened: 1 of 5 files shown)


react/components/Container/Container.css
```diff
@@ -0,0 +1 @@
+.container {}
```
kevinch: Useful?

lbebber: yes, it adds a CSS handle to allow customizing the component via CSS modules with the styles-builder

PR #420, opened 2019-04-17
Group sku variations by variation name

lucasecdb:
Instead of showing all items at once, group them by the variation and show only the first one.

The PR #419 introduced this regretion while removing the sort of the items.

[Workspace](https://lucas--alssports.myvtex.com/under-armour-remix-running-shoes---men-s/p?skuId=1).

#### Types of changes


CHANGELOG.md
```diff
@@ -6,6 +6,8 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+### Fixed
+- Fix variations not being grouped.
```
salesfelipe:
```suggestion
- Variations not being grouped.
```

react/components/SKUSelector/utils/index.js
```diff
@@ -41,6 +41,16 @@ export const parseSku = sku => {
   return result
 }
 
+/**
+ * Group the sku items by the variation name specified
+ */
+export const groupItemsByVariation = (name, items) =>
```
salesfelipe: please, don't be so ramdeee. this is no't js

react/components/SKUSelector/components/SelectorItem.js
```diff
@@ -89,17 +95,4 @@ SelectorItem.propTypes = {
...
-}
-
-const getDiscount = (maxPrice, price) => {
-  let discount = 0
-  if (maxPrice && price) {
-    discount = 1 - price / maxPrice
-  }
-  return discount
-}
+export default memo(SelectorItem)
```
jgfidelis: is it smart to use memo on a component that makes use of the "children" prop? see thread: https://github.com/facebook/react/issues/14463

react/components/SKUSelector/utils/index.js
jgfidelis: I don't  see a problem. maybe post a example of input and output of that function so we can understand and learn what the function should do

lucasecdb: we use jsx daily, and it's *literally* not JavaScript, but we don't have any problems using it, why does this bother you?

react/components/SKUSelector/components/SelectorItem.js
lucasecdb: I just refactored it to a function component, and since it was a pure component before I added this hoc, but good call :+1:

react/components/SKUSelector/utils/index.js
salesfelipe: @jgfidelis already answered for me, what's the point of having a code that I have to ready a 'doc' to understand what it does, using ramda this way you are writing basically a black box.

salesfelipe: Which makes harder to maintain to your peers

lucasecdb:
releasing twice because I forgot to rebase the changes from master 🤦‍♂️

PR #423, opened 2019-04-22
make UserAddress use modal block

jgfidelis:
use `modal` block on UserAddress.

https://address2--delivery.myvtex.com/


react/components/UserAddress/index.js
```diff
@@ -1,135 +1,125 @@
 import PropTypes from 'prop-types'
```
jgfidelis: had to convert this to functional component to use `useChildBlock` hook.

PR #436, opened 2019-05-03
Carousel renders a placeholder when there is no image

estacioneto:

As the title says.

[Infinite content loader when there is no image to be rendered](https://app.clubhouse.io/vtex-dev/story/10296/product-without-image-with-infinite-loading)

[Access the workspace](https://placeholder--invictastores.myvtex.com/impact-cases) and click a product with no image

image

image

#### Types of changes

### DIFF
--- a/react/components/ProductImages/components/Carousel/Loader.js
+++ b/react/components/ProductImages/components/Carousel/Loader.js
@@ -1,40 +1,11 @@
 import ContentLoader from 'react-content-loader'
 import React from 'react'
 
+import ImagePlaceholder from './ImagePlaceholder'
+
 export default props => {
   if (props.slidesAmount === 0) {
-    return (
-      <svg
-        width="100%"
-        height="100%"
-        viewBox="0 0 512 512"
-        fill="none"
-        xmlns="http://www.w3.org/2000/svg"
-      >
-        <rect width="512" height="512" fill="#F2F2F2" />
-        <rect
-          x="183.857"
-          y="180.2"
-          width="144.286"
-          height="150.474"
-          stroke="#CACBCC"
-          strokeWidth="2"
-        />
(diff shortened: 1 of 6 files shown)


react/components/ProductImages/components/Carousel/Loader.js
```diff
@@ -2,6 +2,40 @@ import ContentLoader from 'react-content-loader'
 import React from 'react'
 
 export default props => {
+  if (props.slidesAmount === 0) {
+    return (
+      <svg
```
klzns: I think it would be better to place this SVG in another file.

Does it make sense to be in `store-icons`? Or that's far fetched?

estacioneto: It can be in another file. I thought it would be too much, but if you see it that way, no problem in changing it.
I don't know if it's better to be a `store-icon` because it feels like an image, to me. Really don't know if it's semantically wrong to put it like this.

klzns: About creating another file, I think it would just make it more clearer, since you would have to give it a name.

Example:

```jsx
import ImagePlaceholder from './ImagePlaceholder'

// ...

if (props.slidesAmount === 0) {
  return <ImagePlaceholder />
}
```

klzns: You could also mock this in tests, since the real SVG doesnt make a difference in snapshots

estacioneto: Cool. I'll do it :)

klzns: Than it could appear as something like this in snapshots

```jsx
<div>Mocked Image Placeholder</div>
```

PR #444, opened 2019-05-10
Fix/results click

lbebber:
Fixes bug where clicking on search result items would just close the search results list.

Before: https://samsungar.myvtex.com/ (click on the search, type "gala", and try clicking on an item)
After: https://lbebber--samsungar.myvtex.com/ (click on the search, type "gala", and try clicking on an item)

#### Types of changes

### DIFF
--- a/react/components/SearchBar/components/ResultsList.js
+++ b/react/components/SearchBar/components/ResultsList.js
@@ -16,6 +16,7 @@ const ResultsList = ({
   inputValue,
   closeMenu,
   onClearInput,
+  getItemProps,
 }) => {
   const items = data.autocomplete ? data.autocomplete.itemsReturned : []
   const {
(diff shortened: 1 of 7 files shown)


react/components/SearchBar/components/ResultsList.js
```diff
@@ -97,15 +103,16 @@ const ResultsList = ({
...
+              {...getItemProps({
+                item,
+                onClick: handleItemClick,
+              })}
               {...getLinkProps(item)}
               className={listItemClassNames}
             >
               {item.thumb && (
                 <img
+                  alt=""
```
vitoriaHeliane: Isn't there an value to set the `alt` avoiding set it with an empty string?

lbebber: The linter encourages setting it to an empty string when the image is decorative (this case has the product name right after the image, so I don't think there was much to gain from adding any alt text to the image)

PR #481, opened 2019-06-07
Fix/ie11 (part 1)

lbebber:
Fixed issue that causes syntax error on IE11, due to the use of ES6 features.

#### Types of changes

### DIFF
--- a/react/components/InfoCard/index.js
+++ b/react/components/InfoCard/index.js
@@ -132,7 +132,7 @@ const InfoCard = ({
   )
 
   const textContainerClasses = classNames(
-    `${styles.infoCardTextContainer} flex flex-column`,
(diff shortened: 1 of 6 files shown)


react/components/InfoCard/index.js
```diff
@@ -132,7 +132,7 @@ const InfoCard = ({
   )
 
   const textContainerClasses = classNames(
-    `${styles.infoCardTextContainer} flex flex-column`,
+    `${styles.infoCardTextContainer} flex flex-column mw-100`,
```
GustavoMA: No need to use String Interpolation if you're using `classNames`.
```javascript
const textContainerClasses = classNames(
  styles.infoCardTextContainer,
  'flex flex-column mw-100',
  ...
)
```

PR #486, opened 2019-06-14
Fix/sku iterative

jgfidelis:
https://fidelis--samsungar.myvtex.com/galaxy-s10/p
https://fidelis--alssports.myvtex.com/sitka-hoody-kelvin-lite-optifade/p
https://fidelis--storecomponents.myvtex.com/classic-shoes/p

### DIFF
--- a/react/components/SKUSelector/Wrapper.js
+++ b/react/components/SKUSelector/Wrapper.js
@@ -21,9 +21,9 @@ const useVariations = (skuItems, shouldNotShow) => {
     }
     const variationsNames = Object.keys(variations)
     // Transform set back to array
-    for (const varName of variationsNames) {
-      const set = variations[varName]
-      variations[varName] = Array.from(set)
+    for (const variationName of variationsNames) {
+      const set = variations[variationName]
+      variations[variationName] = Array.from(set)
     }
     return variations
   }, [skuItems, shouldNotShow])
--- a/react/components/SKUSelector/Wrapper.js
+++ b/react/components/SKUSelector/Wrapper.js
@@ -1,12 +1,30 @@
-import React, { useContext } from 'react'
+import React, { useContext, useMemo } from 'react'
 import { ProductContext } from 'vtex.product-context'
 import { pathOr } from 'ramda'
 
 import SKUSelector from './index'
 
+const useVariations = (skuItems, shouldNotShow) => {
+  const variations = useMemo(() => shouldNotShow ? {} : skuItems.reduce((totalAcc, skuItem) => {
+    const indivVariations = skuItem.variations.reduce((indivAcc, currentVariation) => {
+      const { name, values } = currentVariation
+      const value = values[0]
+      const currentArray = totalAcc[name] || []
+      if (!currentArray.includes(value)) {
+        currentArray.push(value)
+      }
+      return { ...indivAcc, [name]: currentArray }
+    }, {})
+    return {
+      ...totalAcc,
+      ...indivVariations,
+    }
+  }, {}), [skuItems, shouldNotShow])
+  return variations
+}
+
 const SKUSelectorWrapper = props => {
   const valuesFromContext = useContext(ProductContext)
-
   const skuItems =
     props.skuItems != null
       ? props.skuItems
@@ -16,24 +34,27 @@ const SKUSelectorWrapper = props => {
     props.skuSelected != null
       ? props.skuSelected
       : valuesFromContext.selectedItem
-
-  if (
-    skuItems.length <= 1 ||
+  
+  const shouldNotShow = skuItems.length <= 1 ||
     !skuSelected ||
     !skuSelected.variations ||
     skuSelected.variations.length === 0
-  ) {
+
+  const variations = useVariations(skuItems, shouldNotShow)
+
+  if (shouldNotShow) {
     return null
   }
 
   return (
     <SKUSelector
-      alwaysShowSecondary={props.alwaysShowSecondary}
       onSKUSelected={props.onSKUSelected}
       skuItems={skuItems}
       skuSelected={skuSelected}
       maxItems={props.maxItems}
       seeMoreLabel={props.seeMoreLabel}
+      variations={variations}
+      hideImpossibleCombinations={props.hideImpossibleCombinations}
     />
   )
 }
--- a/react/components/SKUSelector/Wrapper.js
+++ b/react/components/SKUSelector/Wrapper.js
@@ -14,7 +14,7 @@ const useVariations = (skuItems, shouldNotShow) => {
       for (const currentVariation of skuItem.variations) {
         const { name, values } = currentVariation
         const value = values[0]
-        const currentSet = variations[name] || (new Set())
+        const currentSet = variations[name] || new Set()
         currentSet.add(value)
         variations[name] = currentSet
       }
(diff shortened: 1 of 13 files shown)


react/components/SKUSelector/Wrapper.js
```diff
@@ -1,12 +1,37 @@
...
+      }
+    }
+    const variationsNames = Object.keys(variations)
+    // Transform set back to array
+    for (const varName of variationsNames) {
+      const set = variations[varName]
+      variations[varName] = Array.from(set)
+    }
+    return variations
+  })
```
lbebber: you are using `useMemo`, but are not setting its dependencies. this renders its usage moot afaik

react/components/SKUSelector/components/SKUSelector.js
```diff
@@ -1,69 +1,159 @@
...
+  const selectedNotNull = filter(Boolean, selectedVariations)
+  const selectedCount = Object.keys(selectedNotNull).length
+  if (selectedCount === variationCount && isSelected) {
+    const item = findItemWithSelectedVariations(possibleItems, selectedVariations)
+    return isSkuAvailable(item)
+  }
+  return possibleItems.some(isSkuAvailable)
+}
+
+const getAvailableVariations = ({ variations, selectedVariations, imagesMap, onSelectItemMemo, skuItems, hideImpossibleCombinations }) => {
```
lbebber: :pick:  why `onSelectItemMemo` instead of `onSelectItem`?

lbebber: what is `imagesMap`?

react/components/SKUSelector/components/SKUSelector.js
```diff
@@ -1,69 +1,159 @@
...
-
-  const secondaryOnSelectItem = useCallback(skuId => onSelectSKU(false, skuId), [onSelectSKU])
-  const secondaryCheckSelected = useCallback(sku => sku.itemId === secondaryVariation.value, [secondaryVariation])
+  const [displayVariations, setDisplayVariations] = useState(null)
+  const onSelectItemMemo = useCallback(
+    ({ name, value, skuId, isMainAndImpossible, possibleItems }) => () => onSelectItem({ name, value, skuId, isMainAndImpossible, possibleItems}), 
+    [onSelectItem]
+  )
+  useEffect(() => {
+    const promise =
```
lbebber: ⛏ no need to attribute this to a variable I think? but whatever

lbebber: using await here would be nice I think

react/components/SKUSelector/Wrapper.js
jgfidelis: its a mistake, thanks

react/components/SKUSelector/components/SKUSelector.js
lbebber: I see now that `imagesMap` is commented below, on the propTypes. perhaps add a comment that these arguments are explained below? just a suggestion though

react/components/SKUSelector/components/SelectorItem.js
```diff
@@ -19,15 +19,16 @@ const getDiscount = (maxPrice, price) => {
...
   isSelected = false,
-  children = null,
   maxPrice,
   price,
   onClick,
   isImage,
   variationValue,
+  imageUrl,
+  imageLabel,
+  isFaded,
```
lbebber: ⛏ I think this name should be semantic, because the styling might change, but the meaning should not. something like, `isImpossible` maybe? (since that's the term that was used)

react/components/SKUSelector/index.js
```diff
@@ -1,110 +1,150 @@
...
+    result[varName] = null
+  }
+  return result
+}
+
+/** receives an item and the variations object, returns the selected variations for that item */
+const selectedVariationFromItem = (item, variations) => {
+  const variationNames = Object.keys(variations)
+  const result = {}
+  for (const varName of variationNames) {
```
lbebber: ⛏ please use `variationName` here (no need for shortening + it can be confused with "variable name")

react/components/SKUSelector/index.js
```diff
@@ -1,110 +1,150 @@
...
-  const handleSkuSelection = useCallback((isMainVariation, skuId) => {
-    const sku = parsedItems.find(({ itemId }) => itemId === skuId)
-    const {
-      secondaryVariation: { options },
-    } = buildVariations(sku)
-
-    const isSecondaryPicked =
-      !isMainVariation || (options && options.length === 1)
+    if (selectedSkuId === skuIdToRedirect || isRemoving) {
+      // do nothing
```
lbebber: ⛏ comment on why it does nothing, rather than what is being done (which is clear enough)

react/components/SKUSelector/index.js
```diff
@@ -1,110 +1,150 @@
...
+      return
```
lbebber:
```suggestion
      return null
```

unless I'm missing something—returning void on render causes bugs on react, doesnt it?

react/components/SKUSelector/index.js
```diff
@@ -1,110 +1,150 @@
...
 
     if (onSKUSelected) {
-      onSKUSelected(skuId, isMainVariation, isSecondaryPicked)
+      setSelectedSkuId(skuIdToRedirect)
+      onSKUSelected(skuIdToRedirect)
     } else {
-      redirectToSku(skuId, isMainVariation)
+      if (allSelected || (isColor(variationName))) {
+        setSelectedSkuId(skuIdToRedirect)
+        setTimeout(() => redirectToSku(skuIdToRedirect), 0)
```
lbebber: this is perhaps something you should add a comment explaining why it was done

react/components/SKUSelector/utils/index.js
```diff
@@ -41,38 +30,13 @@ export const stripUrl = url => url.replace(/^https?:/, '')
...
-
-  const variations = sku.variations.map(variation => {
-    result[variation.name] = variation.values[0]
-    return variation.name
-  })
-
-  result.variations = variations
-
+  for(const variation of sku.variations) {
+    result[variation.name] = variation.values[0];
```
lbebber: missing formatting here?

react/components/SKUSelector/utils/index.js
```diff
@@ -96,3 +60,81 @@ export const getMainVariationName = variations => {
...
+export const uniqueOptionToSelect = (items, selectedVariations, isMainAndImpossible) => {
+  const possibleItems = !isMainAndImpossible ? items : findListItemsWithSelectedVariations(items, selectedVariations)
+  const unselected = reject(Boolean, selectedVariations)
+  const unselectedNames = Object.keys(unselected)
+  const availableOptions = buildAvailableVariations(possibleItems, unselectedNames)
+  const variationsWithOne = filter(setValues => setValues.size === 1, availableOptions)
+
+  const variationsNames = Object.keys(variationsWithOne)
+  // Transform set to plain value
+  for (const varName of variationsNames) {
```
lbebber: ⛏ variationName please 😬

react/components/SKUSelector/utils/index.js
```diff
@@ -96,3 +60,81 @@ export const getMainVariationName = variations => {
...
+  return hasAll
+}
+
+const buildAvailableVariations = (items, variationNames) => {
+  const result = {}
+  for (const varName of variationNames) {
+    result[varName] = new Set()
+  }
+  for (const item of items) {
+    for (const varName of variationNames) {
```
lbebber: ⛏ here too

react/components/SKUSelector/index.js
jgfidelis: This is inside the onSelectItem callback, so `return` here is an early exit to a function.

PR #489, opened 2019-06-15
Show search option when no result is found

klzns:
#### Types of changes

### DIFF
--- a/react/components/SearchBar/components/ResultsList.js
+++ b/react/components/SearchBar/components/ResultsList.js
@@ -102,7 +102,7 @@ const ResultsList = ({
                   id="store/search.searchFor"
                   values={{
                     term: (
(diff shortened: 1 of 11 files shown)


react/components/SearchBar/components/ResultsList.js
```diff
@@ -9,10 +10,17 @@ import autocomplete from '../queries/autocomplete.gql'
 
 import styles from '../styles.css'
 
+const WrappedSpinner = () => (
+  <div className="w-100 flex justify-center">
```
jgfidelis: maybe drop w-100 and use flex-grow-1 ?
@lbebber what do you think?

lbebber: sounds good if the parent is a flex element

klzns: It's not

PR #494, opened 2019-06-19
Imports messages from native types

tlgimenes:
This app is currently using native types messages without explicitly declaring their usage. This pr solves this problem by adding these messages to `context.json`

#### Types of changes

### DIFF
--- a/messages/context.json
+++ b/messages/context.json
@@ -152,5 +152,7 @@
   "admin/editor.skuSelector.seeMoreLabel.description": "Description of seeMoreLabel property",
   "vtex.native-types::store/native-types.text": "native types",
   "vtex.native-types::store/native-types.richText": "Rich text field",
-  "vtex.native-types::store/native-types.url": "URL field"
+  "vtex.native-types::store/native-types.url": "URL field",
+  "vtex.address-form::address-form.field.postalCode": "address-form.field.postalCode",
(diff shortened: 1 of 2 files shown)


messages/context.json
```diff
@@ -149,5 +149,10 @@
   "admin/editor.skuSelector.title": "Title of SKU Selector component",
   "admin/editor.skuSelector.description": "Description of SKU Selector component",
   "admin/editor.skuSelector.seeMoreLabel.title": "Title of seeMoreLabel property",
-  "admin/editor.skuSelector.seeMoreLabel.description": "Description of seeMoreLabel property"
+  "admin/editor.skuSelector.seeMoreLabel.description": "Description of seeMoreLabel property",
+  "vtex.native-types::store/native-types.text": "native types",
+  "vtex.native-types::store/native-types.richText": "Rich text field",
+  "vtex.native-types::store/native-types.url": "URL field",
+  "vtex.address-form::address-form.field.postalCode": "address-form.field.postalCode",
```
klzns: Let's discuss this. 

Store-components it's not using this message implicitly. Store-components it's using a component from the vtex.address-form app and this component it's using its own message. 

The point is, this message is an implementation detail of the app address-form.

tlgimenes:
In fact, we don't need to add these messages in this context. This import should be resolved by our framework in `vtex.pages-graphql`

PR #497, opened 2019-06-19
Improve readme availability-subscriber

denisebiscaro:

Improved read-me of availability-subscriber component.
Improved default text of component to make it less misleading.

Users usually believe that this component warn the users when a product become available. However it only makes a list on Master Data of users that subscribe to that product.

---

---

#### Types of changes
- [X] Bug fix (a non-breaking change which fixes an issue)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,8 +8,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ## [Unreleased]
 
 ### Fixed
-- Fixed read me of Availability subscriber. Explain that it doesn't warn users.
-- Adapted default text of component to make it less misleading.
+- Read me of Availability subscriber. Explain that it doesn't warn users.
+- Default text of component to make it less misleading.
 
 ## [3.47.5] - 2019-06-27
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
+### Fixed
+- Fixed read me of Availability subscriber. Explain that it doesn't warn users.
+- Adapted default text of component to make it less misleading.
+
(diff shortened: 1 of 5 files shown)


CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
+### Fixed
+- Fixed read me of Availability subscriber. Explain that it doesn't warn users.
```
kevinch: No need to repeat the verb "Fixed" since it's the title of the section already.

CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
...
+- Adapted default text of component to make it less misleading.
```
kevinch:
```suggestion
- Default text of component to make it less misleading.
```

kevinch: Same here.

messages/en.json
```diff
@@ -39,7 +39,7 @@
   "store/buybutton.see-cart": "View cart",
   "store/technicalspecifications.title": "Technical Specifications",
   "store/availability-subscriber.title": "This product isn't available right now",
-  "store/availability-subscriber.subscribe-label": "Let me know when this product is available again",
+  "store/availability-subscriber.subscribe-label": "I want to know when this product is available again",
```
kevinch: Maybe "...will be available again"?

denisebiscaro: Not sure if I agree, because it sounds like the user wants to know the `date` when it will become available. 

But I agree that it can be improved, so what do you think about changing it to `I want to know when this product become available`.

Or maybe we could change the whole meaning (then I could change on the other languages as well), and change it to someting like `Subscribe to this product` or `Follow this product`. 

What do you think @kevinch ?

kevinch: I like your suggestion "I want to know when this product become available", thought of "Let me know know when this product becomes available.".

PR #517, opened 2019-07-04
fix for one item

jgfidelis:
https://fidelis--alssports.myvtex.com/tagua-holster-4-in-1-sw-mp-compact/p?skuId=11565


react/components/SKUSelector/Wrapper.js
```diff
@@ -43,7 +43,7 @@ const SKUSelectorWrapper = props => {
       : valuesFromContext.selectedItem
 
   const shouldNotShow =
-    skuItems.length <= 1 ||
+    skuItems.length < 1 ||
```
klzns:
```suggestion
    skuItems.length === 0 ||
```

PR #554, opened 2019-08-19
Handle user adding a duplicate item to the minicart

victorhmp:

Add new message to be shown when the user tries to add an item to the minicart that is already there.

Also, update ESLint version and make necessary adjustments so that `yarn lint` works :)

[Workspace](https://victorhmp--storecomponents.myvtex.com/)

#### Checklist/Reminders

- [ ] Updated `README.md`.
- [x] Updated `CHANGELOG.md`.

#### Type of changes

✔️ | Type of Change
---|---
_ | Bug fix 
✔️ | New feature 
_ | Breaking change 
✔️ | Technical improvements


react/components/BuyButton/index.js
```diff
@@ -64,14 +65,25 @@ export const BuyButton = ({
   const translateMessage = useCallback(id => intl.formatMessage({ id: id }), [
     intl,
   ])
+  const orderFormItems =
```
jgfidelis: just use ramda's path...

react/components/BuyButton/index.js
```diff
@@ -116,20 +128,36 @@ export const BuyButton = ({
...
-          ))
+      const addedItem =
+        linkStateItems &&
+        skuItems.filter(
+          skuItem => !!linkStateItems.find(({ id }) => id === skuItem.skuId)
+        )
+
+      const foundItem =
+        orderFormItems &&
+        orderFormItems.filter(item => item.id === addedItem[0].skuId).length > 0
```
jgfidelis: why only item[0]?

react/components/BuyButton/index.js
```diff
@@ -64,14 +65,25 @@ export const BuyButton = ({
   const translateMessage = useCallback(id => intl.formatMessage({ id: id }), [
     intl,
   ])
+  const orderFormItems =
+    orderFormContext &&
+    orderFormContext.orderForm &&
+    orderFormContext.orderForm.items
+  
+  const resolveToastMessage = (success, isNewItem) => {
+    if (!success) return translateMessage(CONSTANTS.ERROR_MESSAGE_ID)
```
jgfidelis: just check if isNewItem here and early return

react/components/BuyButton/index.js
victorhmp: Yeeeah... in this case should be better :)

react/components/BuyButton/index.js
victorhmp: Wow! That should probably not be a `[0]`... but I don't know in which case the `addedItem` would be an array with length greater than 1. Would that be the case with Assembly Options?

victorhmp:
```
const addedItem =
        linkStateItems &&
        skuItems.filter(
          skuItem => !!linkStateItems.find(({ id }) => id === skuItem.skuId)
        )
```

This is how `addedItem` is defined. For it to contain more than one item, it would require `linkStateItems` to contain two items with the same skuId, and I'm aaalmost sure that should not happen.

jgfidelis: no problem, [0] is ok

react/components/BuyButton/index.js
```diff
@@ -116,20 +126,36 @@ export const BuyButton = ({
...
+        )
+
+      const foundItem =
+        orderFormItems &&
+        orderFormItems.filter(item => item.id === addedItem[0].skuId).length > 0
+
+      success = success || addedItem
 
-      showToastMessage = () => toastMessage(success && success.length >= 1)
+      showToastMessage = foundItem
```
jgfidelis: this should be: `() =>
            toastMessage({
              success: success && success.length >= 1,
              isNewItem: ! foundItem,
            })`

victorhmp: Doone!

PR #559, opened 2019-08-21
Prevent addToCart while orderForm is loading

gustavorosolem:

If the user clicks before the order form is loaded, the value will be null and the product will not be added in the cart.
```
const variables = {
          orderFormId: orderFormContext.orderForm.orderFormId,
          items: skuItems.map(item => ({
            id: item.skuId,
            seller: item.seller,
            options: item.options,
            quantity: item.quantity,
          })),
        }
```
You can notice more when in `isOneClickBuy`.

[Workspace](https://gustavo--qastore.mygocommerce.com/produto-celular/p)

#### Checklist/Reminders

- [ ] Updated `README.md`.
- [x] Updated `CHANGELOG.md`.

#### Type of changes

✔️ | Type of Change
---|---
✔️ | Bug fix 
_ | New feature 
_ | Breaking change 
_ | Technical improvements

### DIFF
--- a/react/components/BuyButton/index.js
+++ b/react/components/BuyButton/index.js
@@ -77,7 +77,7 @@ export const BuyButton = ({
       ? translateMessage(CONSTANTS.SUCCESS_MESSAGE_ID)
       : translateMessage(CONSTANTS.OFFLINE_BUY_MESSAGE_ID)
 
-    return checkForOffline 
+    return checkForOffline
   }
 
   const toastMessage = ({ success, isNewItem }) => {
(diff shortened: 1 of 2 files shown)


react/components/BuyButton/index.js
```diff
@@ -173,7 +173,7 @@ export const BuyButton = ({
       ) : (
         <Button
           block={large}
-          disabled={disabled || !available}
+          disabled={disabled || !available || orderFormContext.loading}
```
klzns: I think it's better to active the prop `isLoading` instead

gustavorosolem: @klzns I agree that sounds better, but the UX will be weird.
When load the page, the user will see the button spinning, even thought he don't clicked, I don't think this behavior is expected.
What do you think?

PR #584, opened 2019-09-18
Send InputValues to GraphQL.

klzns:
Also added some tests to the assemblyOption module.

Linked workspace:
https://breno--storecomponents.myvtex.com/star-color-top/p


react/components/BuyButton/assemblyUtils.ts
```diff
@@ -0,0 +1,187 @@
...
+          seller: item.seller,
+          ...(childrenOptions && childrenOptions.length > 0
+            ? { options: childrenOptions }
+            : {}),
+        })
+      }
+    }
+  }
+
+  const assemblyInputValuesKeys: GroupId[] = Object.keys(inputValues)
```
klzns: From here to 177 was the thing that I had to add to make it work.

react/components/BuyButton/Wrapper.js
```diff
@@ -178,6 +177,7 @@ EnhancedBuyButton.mapCatalogItemToCart = function mapCatalogItemToCart({
         imageUrl: path(['images', '0', 'imageUrl'], selectedItem),
         ...transformAssemblyOptions(
           path(['items'], assemblyOptions),
+          path(['inputValues'], assemblyOptions),
```
jgfidelis: if this returns null it will break the buy button, after your changes, are you comfortable enough with this?

(if its null will not get into the {} fallback)

klzns: Yepp

PR #627, opened 2019-11-01
Add items to cart validating with id and seller

grupo-exito-ecommerce:
Currently we working a functionality that allows the user to see the "additional offers" of the same product, in other words means to see all the offers of the different sellers associated with a product, with this possibility the user can add the same product to the cart but sold by a different "seller". We try to perform this functionality using the BuyButton of the store-component, but when you try to add the product with another seller the buy button notice that the product is already added to the cart.

We note that the BuyButton uses the “addToCart” mutation (minicart\react\localState\index.js) to add to the cart, here you will find the new items that must be added in the local state should not have been added recently, such validation is performed using the "id" of the item, which is insufficient and does not allow adding that item with different vendors. 

The main change proposed was made in the [Minicart](https://github.com/vtex-apps/minicart/pull/183) project, but we note that in the BuyButton component they validate that the new item was added to the cart using a validation with "skuId", so we add the id of the "seller" in validation.

screencapture-chrome-extension-fdpohaocaechififmbbbbbknoalclacl-editor-html-2019-11-01-14_08_56

#### Type of changes

✔️ | Type of Change
---|---
✔️| Bug fix 
_ | New feature 
_ | Breaking change 
_ | Technical improvements

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ### Added
 - Support to block class.
 
+### Fixed
+- Verify that the item has been added to the cart by filtering `skuId` and `seller` in the BuyButton.
+
 ## [3.85.0] - 2019-11-12
 ### Added
 - Option to edit `visibleSpecifications` and `hiddenSpecifications` in Site Editor.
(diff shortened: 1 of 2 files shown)


klzns:
Thanks for the PR! We will merge them until the end of the week.

PR #643, opened 2019-11-14
Use icons from store-icons as blocks (SearchBar)

victorhmp:

Allow icons from `vtex.store-icons` to be passed as blocks to the `search-bar` block. That way users can customize each individual icon.

See how weird the icons from the SearchBar look:
[Workspace](https://storeicons--storecomponents.myvtex.com/)

#### Checklist/Reminders

- [x] Updated `README.md`.
- [x] Updated `CHANGELOG.md`.

#### Type of changes

✔️ | Type of Change
---|---
_ | Bug fix 
✔️ | New feature 
_ | Breaking change 
_ | Technical improvements 

#### Notes

This depends on: https://github.com/vtex-apps/store-icons/pull/37


docs/SearchBar.md
```diff
@@ -20,16 +20,17 @@ Then, add `search-bar` block into your app theme, as we do in our [Store Header]
...
+| `autoFocus`           | `Boolean`                               | Identify if the search input should autofocus or not                                                                                              | -             |
+| `maxWidth`            | `Number` \| `String`                    | Max width of the search bar                                                                                                                       | -             |
+| `customSearchPageUrl` | `string`                                | Template for a custom url. It can have a substring `${term}` used as placeholder to interpolate the searched term. (e.g. `/search?query=${term}`) | -             |
+| ~`iconClasses`~       | `String`                                | **DEPRECATED** ~Custom classes for the search icon~ Use the CSS handle `searchBarIcon`.                                                           | -             |
+| ~`iconsProps`~        | `{ "viewBox": String, "size": Number }` | Props to be passed down to the icons from `store-icons` being used by SearchBar.                                                                  | -             |
```
klzns:
```suggestion
| `iconsProps`        | `{ "viewBox": String, "size": Number }` | Props to be passed down to the icons from `store-icons` being used by SearchBar.                                                                  | -             |
```

react/components/SearchBar/components/AutocompleteInput.js
```diff
@@ -107,6 +108,12 @@ AutocompleteInput.propTypes = {
   iconBlockClass: PropTypes.string,
   /** Identify if the search input should autofocus or not */
   autoFocus: PropTypes.bool,
+  iconsProps: PropTypes.shape({
```
klzns: 🤔 should we create two separate props for each icon?

victorhmp: I thought about it... but I think it's better to have a single prop for consistency.
If you set a viewBox for one of the icons, for example, you're setting it for the others as well.

lbebber: > If you set a viewBox for one of the icons, for example, you're setting it for the others as well
not sure if I get this.

is this behaviour unavoidable? If so, it of course makes sense for both icons to share the same prop, but I don't think this is the ideal behaviour.

docs/SearchBar.md
```diff
@@ -20,16 +20,16 @@ Then, add `search-bar` block into your app theme, as we do in our [Store Header]
...
-| `autoFocus`           | `Boolean`            | Identify if the search input should autofocus or not                                                                                              | -             |
-| `maxWidth`            | `Number` \| `String` | Max width of the search bar                                                                                                                       | -             |
-| `customSearchPageUrl` | `string`             | Template for a custom url. It can have a substring `${term}` used as placeholder to interpolate the searched term. (e.g. `/search?query=${term}`) | -             |
-| ~`iconClasses`~       | `String`             | **DEPRECATED** ~Custom classes for the search icon~ Use the CSS handle `searchBarIcon`.                                                           | -             |
+| Prop name             | Type                                    | Description                                                                                                                                       | Default value |
```
Klynger: This is so hard to maintain and after some time you can't even see the table formatted in the vscode

victorhmp: I know! We need a better way to show the list of available props

PR #654, opened 2019-12-02
Make BuyButton redirect to the appropriate cart

marcoskwkm:

We're developing a new Checkout within IO, and along with it we built a new Cart. It is located in a different path (`/cart` instead of `/checkout/#/cart`), so we need to change this component to redirect to the new cart **when `vtex.checkout@1.x` is installed in the account**. To do so this PR makes a query to `apps-graphql` to find out which major of `vtex.checkout` is installed in the account and defines the checkout URL accordingly.

[Workspace with major 0](https://buybutton0--checkoutio.myvtex.com/teste-so-delivery-copy-256--copy-257--copy-261--copy-262--copy-263--copy-264-/p)
[Workspace with major 1](https://buybutton1--checkoutio.myvtex.com/teste-so-delivery-copy-256--copy-257--copy-261--copy-262--copy-263--copy-264-/p)

After clicking "Add to cart", the first workspace should take you to `/checkout/#/cart`, while the second takes you to `/cart`. Also, the "View cart" button in the toast that is shown after clicking the button should point to the same URL.

#### Checklist/Reminders

- [ ] Updated `README.md`.
- [x] Updated `CHANGELOG.md`.
- [x] [Linked this PR to a Clubhouse story (if applicable)](https://app.clubhouse.io/vtex/story/27290/resolver-direcionamento-para-o-novo-carrinho-no-minicart-e-no-botão-de-adicionar-ao-carrinho-página-de-produto).

#### Type of changes

✔️ | Type of Change
---|---
✔️ | Bug fix 
✔️ | New feature 
_ | Breaking change 
_ | Technical improvements 

#### Notes

This **should not be merged** until we make sure no account in production has `vtex.checkout@1.x` installed.

### DIFF
--- a/react/components/BuyButton/Wrapper.js
+++ b/react/components/BuyButton/Wrapper.js
@@ -12,12 +12,18 @@ import { BuyButton } from './index'
 import { transformAssemblyOptions, sumAssembliesPrice } from './assemblyUtils'
 import addToCartMutation from './mutations/addToCart.gql'
 import setOpenMinicartMutation from './mutations/setOpenMinicart.gql'
+import installedApp from './queries/installedApp.gql'
 
 const MESSAGE_CSS_HANDLES = [
   'buyButtonText',
   'buttonDataContainer',
   'buttonItemsPrice',
 ]
+
+const CHECKOUT_URL = {
+  V0: '/checkout/#/cart',
+  V1: '/cart',
+}
 
 const BuyButtonMessage = ({ showItemsPrice, skuItems }) => {
   const handles = useCssHandles(MESSAGE_CSS_HANDLES)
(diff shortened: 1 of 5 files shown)


react/components/BuyButton/Wrapper.js
```diff
@@ -122,6 +127,13 @@ const BuyButtonWrapper = ({
       ? propDisabled
       : !areAssemblyGroupsValid
 
+  const version =
+    checkoutVersion &&
+    checkoutVersion.installedApp &&
+    checkoutVersion.installedApp.version
+  const checkoutUrl =
+    version && version.split('.')[0] === '1' ? CHECKOUT_URL.V1 : CHECKOUT_URL.V0
```
darlenedms: Wouldn't be better apply for any version greater than `0`? It's only require a reorder:

```suggestion
    version && version.split('.')[0] === '0' ? CHECKOUT_URL.V0 : CHECKOUT_URL.V1
```

marcoskwkm: My concern with this change is that if the GraphQL query fails for some reason, it will redirect to the new cart, which might even exist. The way it is written right now at least it will "fallback" to the current cart.

darlenedms: Makes sense, in this case what do you think about `parseInt` the value and check for greater than `0`?

```js
  const version =
    checkoutVersion &&
    checkoutVersion.installedApp &&
    checkoutVersion.installedApp.version &&
    parseInt(checkoutVersion.installedApp.version)
  const checkoutUrl =
    version && version.split('.')[0] > 0 ? CHECKOUT_URL.V1 : CHECKOUT_URL.V0
```

PR #691, opened 2020-01-14
Show savings as percentage prop added

JNussens:
Show savings as percentage prop added

Show savings as percentage prop added

1. Make `savingsAsPercentage` prop `true` on the `product-price` block
2. Check the product page to se if it renders as a percentage discount

[Workspace](url)
https://dev--miriadeit.myvtex.com/prodotto-4/p
#### Checklist/Reminders

- [x] Updated `README.md`.
- [x] Updated `CHANGELOG.md`.

#### Type of changes

✔️ | Type of Change
---|---
_ | Bug fix 
✔️| New feature 
_ | Breaking change 
_ | Technical improvements


CHANGELOG.md
```diff
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
+- Added `savingsAsPercentage` prop
```
klzns:
```suggestion
### Added
- `savingsAsPercentage` prop
```

react/components/ProductPrice/index.js
```diff
@@ -155,21 +156,32 @@ const ProductPrice = (props, context) => {
...
+            />}
+          {savingsAsPercentage && (listPrice - sellingPrice > 0) &&
+            <div
+              className={classNames(
+                productPrice.listPriceLabel,
+                handles.price_listPriceLabel,
+                'dib ph2 t-small-ns t-mini'
+              )}
+            >
+              {((listPrice - sellingPrice) / 1000) * 100}%
```
klzns: Percentages varies on how it is rendered based on locale. You can read more about this here: https://phrase.com/blog/posts/number-localization/#toc_5

To make it global, use FormattedNumber from `react-intl`:
```suggestion
              <FormattedNumber value={((listPrice - sellingPrice) / 1000) * 100} style="percent" />
```

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [3.98.0] - 2020-01-02
 ### Added
+- `savingsAsPercentage` prop
```
klzns: The PR changes should be under `[Unreleased]`.

Example:

```md
## [Unreleased]

### Added
- `savingsAsPercentage` prop
```

react/components/ProductPrice/index.js
```diff
@@ -143,7 +144,7 @@ const ProductPrice = (props, context) => {
             handles.price_listPriceContainer
           )}
         >
-          {showLabels && (
+          {showLabels && !savingsAsPercentage && (
```
klzns: Why is that necessary?

react/components/ProductPrice/index.js
```diff
@@ -155,21 +156,35 @@ const ProductPrice = (props, context) => {
...
-              listPriceRangeClass,
-              handles.price_listPriceRange,
-            )}
-            singleContainerClasses={classNames(
-              productPrice.listPriceValue,
-              listPriceClass,
-              handles.price_listPrice
-            )}
-          />
+          {!savingsAsPercentage &&
```
klzns: Same thing here

react/components/ProductPrice/index.js
JNussens: Because `savingsAsPercentage` toggles between the default way of showing savings *OR* show savings as percentage. It should not render both ways.

react/components/ProductPrice/index.js
JNussens: Because `savingsAsPercentage` toggles between the default way of showing savings *OR* show savings as percentage. It should not render both ways.

CHANGELOG.md
JNussens: Like so?
```
## [Unreleased]

### Added
  - `savingsAsPercentage` prop

## [3.98.0] - 2020-01-02
### Added
- `showValueForVariation` to `SKUSelector`.
```

klzns: Yes, just mind the extra spaces before the `-`

JNussens: Done!

klzns:
Right now this changes are affecting the list price and the savings. It should only target the savings. 

I did some changes to your code, but I can't push it to your branch. Can you allow that please?
https://help.github.com/pt/github/collaborating-with-issues-and-pull-requests/allowing-changes-to-a-pull-request-branch-created-from-a-fork

I changed some things in this PR and made it to render this instead:

Showing Price | Showing Percentage
---|---
image | image


Also just noticed that the calculation is wrong

JNussens:
> Right now this changes are affecting the list price and the savings. It should only target the savings.
> 
> I did some changes to your code, but I can't push it to your branch. Can you allow that please?
> https://help.github.com/pt/github/collaborating-with-issues-and-pull-requests/allowing-changes-to-a-pull-request-branch-created-from-a-fork
> 
> I changed some things in this PR and made it to render this instead:
> 
> Showing Price	Showing Percentage
> image	image
> Also just noticed that the calculation is wrong

I sent you an invite to collaborate. I think now you will have access, can you please confirm? Thanks!

klzns:
@JNussens just push it. Let me know what do you think. 

Also, we need to fix the calculation of the percentage.

klzns:
Just explaining what I did:

We are avoiding creating boolean props. We think that they are very limited and do not scale well.

So I created a new prop called `savingsDisplay` which receives an enum. Props with enum are much more flexible, since we can add new values at will without creating a new prop.

For example, in this case I can imagine a new value for `savingsDisplay` like `biggest-perceived-value` which will render percentage or price, depending which one has bigger numerical value.

JNussens:
@klzns I left most of what you done, just fixed the condition and where the magic happens. This PR was suppose to control the list price section in specific (I think I made a mistake and we all got confused). I just changed the position where our percentage logic shall render.

This is the default result:
image

This is the expected result (using percentage):
image

Now I think everything is correct and suitable for a merge. Let me know if anything needs a fix.

Thanks!

klzns:
I don't understand this design 🤔 It doesn't make much sense too me.
I think this behaviour is very specific to this store. 

Which led me to think that our price component is not well designed or flexible.

I created an issue with a spec of what will be implemented:
https://github.com/vtex-apps/store-discussion/issues/175

klzns:
@JNussens Now it's possible to achieve this with our new app product-price: https://github.com/vtex-apps/product-price

Thanks for bringing the issues to the surface that led us create this new app!

PR #727, opened 2020-03-17
Addition of markdown support at infocard component

marcosewbank:

This feature adds the support to use markdown bold and italic at headline and subhead at Infocard component.

_Update_

This feature's implementation was updated so that the markdown support is **not** enabled by default but through the use of a new `textMode` prop.

You can test it on:

* https://infocard--acctglobal.myvtex.com/

_Update_

You can test the update version here:

https://victorhmp--storecomponents.myvtex.com

Elements structure example: http://i.imgur.com/tKa36Kc.png

#### Checklist/Reminders

- [x] Updated `README.md`.
- [x] Updated `CHANGELOG.md`.

#### Type of changes

✔️ | Type of Change
---|---
_ | Bug fix 
✔️ | New feature 
_ | Breaking change 
_ | Technical improvements 

#### Notes

Closes https://github.com/vtex-apps/store-discussion/issues/236


CHANGELOG.md
```diff
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
+- Addition of markdown support for headline and subhead of infocard component.
```
kaisermann: Please adhere to the keep a changelog format: https://keepachangelog.com/en/1.0.0/

```suggestion
### Added
- Markdown support for headline and subhead of infocard component. 
```

marcosewbank:
@lbebber I belive that by doing it would imply that an user could add tags like  and <iframe> that are allowed in the rich-text component thus providing very weird behaviors to the infocard

kaisermann:
@marcosewbank The decision of using an img/iframe in the info card would be up to the developer. Being able to do something doesn't mean you should. As @lbebber said, we already import `marked` in the `rich-text` app, adding it here would lead to extra weight (at least 8kb) to pages.

marcosewbank:
Understood @kaisermann @lbebber , i made the changes! 
thank you for the amazing tips :smiley:

react/components/InfoCard/index.js
```diff
@@ -1,12 +1,13 @@
...
 import { bool, string, oneOf } from 'prop-types'
+import React, { memo, useState, useEffect, useRef } from 'react'
+import { Renderer } from 'marked'
 import { values } from 'ramda'
-import React, { memo, useMemo } from 'react'
 import { injectIntl, intlShape } from 'react-intl'
 import { useRuntime } from 'vtex.render-runtime'
 import { formatIOMessage } from 'vtex.native-types'
 import { useCssHandles } from 'vtex.css-handles'
+import RichText from 'vtex.rich-text/index'
```
kaisermann:
```suggestion
import RichText from 'vtex.rich-text'
```

react/components/InfoCard/index.js
```diff
@@ -1,8 +1,9 @@
 import classNames from 'classnames'
 import insane from 'insane'
 import { bool, string, oneOf } from 'prop-types'
+import React, { memo, useMemo, useState, useEffect, useRef } from 'react'
+import marked, { Renderer } from 'marked'
```
kaisermann:
```suggestion
```

react/components/InfoCard/index.js
```diff
@@ -98,6 +100,23 @@ const InfoCard = ({
...
+    if (!isMounted) {
+      renderer.current = new Renderer()
+      renderer.current.paragraph = text =>
+        `<p class="lh-copy ${handles.infoCardParagraph}">${text}</p>`
+      renderer.current.strong = text =>
+        `<span class="b ${handles.infoCardStrong}">${text}</span>`
+      renderer.current.em = text =>
+        `<span class="i ${handles.infoCardItalic}">${text}</span>`
+    }
+
```
kaisermann:
```suggestion
```

kaisermann: Is this still needed?

react/package.json
```diff
@@ -32,7 +32,8 @@
     "recompose": "^0.30.0",
     "slick-carousel": "^1.8.1",
     "slugify": "^1.3.4",
-    "swiper": "github:vtex/swiper#4.5.1"
+    "swiper": "github:vtex/swiper#4.5.1",
+    "marked": "^0.6.1"
```
kaisermann:
```suggestion
    "swiper": "github:vtex/swiper#4.5.1"
```

react/components/InfoCard/index.js
```diff
@@ -98,6 +90,23 @@ const InfoCard = ({
...
+  if (!isMounted) {
+    renderer.current = new Renderer()
+    renderer.current.paragraph = text =>
+      `<p class="lh-copy ${handles.infoCardParagraph}">${text}</p>`
+    renderer.current.strong = text =>
+      `<span class="b ${handles.infoCardStrong}">${text}</span>`
+    renderer.current.em = text =>
+      `<span class="i ${handles.infoCardItalic}">${text}</span>`
+  }
+
```
kaisermann: Is this still used?

```suggestion
```

react/components/InfoCard/index.js
```diff
@@ -55,25 +56,16 @@ const safelyGetToken = (tokenMap, valueWanted, propName) =>
...
 const CSS_HANDLES = [
   'infoCardContainer',
   'infoCardTextContainer',
   'infoCardHeadline',
   'infoCardSubhead',
   'infoCardImageContainer',
   'infoCardImage',
+  'infoCardParagraph',
+  'infoCardStrong',
+  'infoCardItalic',
```
kaisermann: Are these still used?

```suggestion
```

react/components/InfoCard/index.js
marcosewbank: Hello @kaisermann, by removing the `/index` on `RichText` import will cause this error: https://i.imgur.com/h4UBjoJ.png

lbebber:
run `yarn test` on the `react` folder to check the failing tests

marcosewbank:
> run `yarn test` on the `react` folder to check the failing tests

I runned the test but an error is appearing at the RichText import. If I don't import using this method the app does not even link :(

with the index import yarn test: http://i.imgur.com/OVgJ2UU.png
without the index import error: http://i.imgur.com/ESOW88X.png

victorhmp:
> @lbebber
> 
> > run `yarn test` on the `react` folder to check the failing tests
> 
> I runned the test but an error is appearing at the RichText import. If I don't import using this method the app does not even link :(
> 
> with the index import yarn test: http://i.imgur.com/OVgJ2UU.png
> without the index import error: http://i.imgur.com/ESOW88X.png

Hi @marcosewbank !
I know this is weird, but the error you're seeing is due to the lack of a mock implementation for the `vtex.rich-text` module, as @lbebber pointed out. This happens because `vtex.rich-text` is not an NPM module, so when these tests are being executed by Jest it won't find the module to use, then throws an error. This exactly the same error you would get if you tried to use a module in any JS project without first installing it using `npm` or `yarn`. The `vtex.rich-text` module and all other modules that are actually VTEX IO apps will not be available to Jest or Travis when trying to execute the tests. Does it make more sense now? It's not that your import is wrong, it's just that modules that cannot be installed using `npm` or `yarn` will break the test setup if you don't provide a mock implementation for them.

In this case, the mock should be fairly simple, creating a `vtex.rich-text.js` file inside the `__mocks__` folder should do! Just make sure you export a `<RichText />` component inside that file and the import should work :)
If you're not familiar with how these mocks will be used in your tests, take a look at https://jestjs.io/docs/en/manual-mocks .

react/components/InfoCard/index.js
Klynger: Hi @marcosewbank, did you try doing this?
```suggestion
import { RichText } from 'vtex.rich-text'
```

marcosewbank: Yes @Klynger, i've tryed but it didn't work :cry: 
the error: https://i.imgur.com/8XfqNZL.png
btw, thank you for the help :smile:

marcosewbank:
> In this case, the mock should be fairly simple, creating a `vtex.rich-text.js` file inside the `__mocks__` folder should do! Just make sure you export a `<RichText />` component inside that file and the import should work :)
> If you're not familiar with how these mocks will be used in your tests, take a look at

@victorhmp I don't have any experience with `mock`, i made this but didn't work, can you help me please

Code: https://i.imgur.com/Jc45Q95.png

victorhmp:
> > In this case, the mock should be fairly simple, creating a `vtex.rich-text.js` file inside the `__mocks__` folder should do! Just make sure you export a `<RichText />` component inside that file and the import should work :)
> > If you're not familiar with how these mocks will be used in your tests, take a look at
> 
> @victorhmp I don't have any experience with `mock`, I made this but didn't work, can you help me, please
> 
> Code: https://i.imgur.com/Jc45Q95.png

Sure! I'll fix the tests and push to this branch so that we can release this :)

I'm just not sure about one tiny detail... if the generated HTML structure didn't change. If it did change, this would result in a breaking change and we should put this feature behind a prop to avoid unexpected changes to layouts already implemented.

victorhmp:
@marcosewbank Just to keep you update on this, I fixed the mocked test, and the tests still fail.
And this is expected... unfortunately this solution changes the resulting HTML structure, turning this into a breaking-change. Also, the `<RichText />` component does nothing with the `classNames` it receives, which means that the handles `infoCardHeadline` and `infoCardSubhead` are made useless by this change since they would never be applied.

What this means is that we can't use this solution as the default one. We should probably toggle this feature using a new prop. As for the CSS handles, maybe a warning in the documentation should do... 
Anyway, I'll change this to be toggled by a new prop, is that OK? What do you think about it?

marcosewbank:
Awesome, thanks for the help @victorhmp  :smile:

react/components/InfoCard/index.js
```diff
@@ -182,28 +185,35 @@ const InfoCard = ({
            * context and inserts a regular image, relying on render runtime
            * to actually make it lazy.
            * In the future, it should be considered to always do this instead,
-           * making this logic simpler, but one must be aware of pontential
+           * making this logic simpler, but one must be aware of potential
```
lbebber, line 188: nice

victorhmp:
Merging this even thought lint is failing because this PR is not responsible for those issues.

PR #766, opened 2020-05-13
Update BuyButton.md

guerreirobeatriz:
Updated the Buy Button documentation, adding a disclaimer regarding the relationship with Minicart v1/v2.

### DIFF
--- a/docs/BuyButton.md
+++ b/docs/BuyButton.md
@@ -2,7 +2,13 @@
 
 # Buy Button
 
-The `BuyButton` block allows users to **add products in the minicart**.
+<!-- ALL-CONTRIBUTORS-BADGE:START - Do not remove or modify this section -->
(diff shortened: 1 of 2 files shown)


docs/BuyButton.md
```diff
@@ -2,21 +2,27 @@
...
-```json
+```diff
   "dependencies": {
-    "vtex.store-components": "3.x"
++   "vtex.store-components": "3.x"
   }
 ```
 
-2. Add the `buy-button` block to any block bellow `store.product` (Product template). For example:
+2. Add the `buy-button` block to other blocks from your theme according to your needs. In the example below, the block is added to the `flex-layout.row` block from the `store.product` template:
```
iaronaraujo, line 25: This passes the idea that it will work in _any_ block, but I believe it really needs to be bellow `store.product` for it to work!

PR #777, opened 2020-05-28
CSS specificationItemProperty + specification name

BeatrizMiranda:
Make it possible to change each of the products specifications by adding a different class for each one 

https://develop--bitofbritain.myvtex.com/rambo-micklem-competion-bridle-dark-havana-sbac9d-nab0/p on the Product Specifications tab

#### Checklist/Reminders

- [x] Updated `README.md`.
- [x] Updated `CHANGELOG.md`.

#### Type of changes

✔️ | Type of Change
---|---
_ | Bug fix 
✔️ | New feature 
_ | Breaking change 
_ | Technical improvements


react/components/ProductSpecifications/index.js
```diff
@@ -122,12 +122,26 @@ const ProductSpecifications = ({
...
             <td
-              className={`${handles.specificationItemProperty} w-50 b--muted-4 bb pa5`}
+              className={`${
+                handles.specificationItemProperty
+              } ${handles.specificationItemProperty +
+                '-' +
+                specification.property.replace(
+                  /\W+/g,
+                  '_'
+                )} w-50 b--muted-4 bb pa5`}
```
victorhmp, line 132: Instead of creating new handles manually, we recommend you use the `applyModifiers` function from `vtex.css-handles`. Here is the documentation for it: https://github.com/vtex-apps/css-handles , there's an example in the README file.

Basically this function will take care of adding the `specification.property` value to the original CSS handle for you, much easier! 😄

react/components/ProductSpecifications/index.js
```diff
@@ -122,12 +122,26 @@ const ProductSpecifications = ({
...
             <td
-              className={`${handles.specificationItemSpecifications} w-50 b--muted-4 bb pa5`}
+              className={`${
+                handles.specificationItemSpecifications
+              } ${handles.specificationItemProperty +
+                '-' +
+                specification.specifications.replace(
+                  /\W+/g,
+                  '_'
+                )} w-50 b--muted-4 bb pa5`}
```
victorhmp, line 144: Same comment from above.

react/components/ProductSpecifications/index.js
BeatrizMiranda: Yes much easier!!

BeatrizMiranda:
> The regex is adding a `_` at the end of the selector, I think it would be better to remove it
> image

I put that because my clients add a ":" on the specification name, so this regex change all the accents and different characters to "_" make it possible to create a CSS selector 
independent of the name

Klynger:
> > The regex is adding a `_` at the end of the selector, I think it would be better to remove it
> > image
> 
> I put that because my clients add a ":" on the specification name, so this regex change all the accents and different characters to "_" make it possible to create a CSS selector
> independent of the name

But after that, you could apply to every handle this regex `handle.replace(/_$/, '')`

CHANGELOG.md
```diff
@@ -7,8 +7,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
```
victorhmp, line 9:
```suggestion
```

victorhmp: @BeatrizMiranda Oh! Just don't forget to apply this suggestion ;)

docs/ProductSpecifications.md
```diff
@@ -51,14 +51,15 @@
...
+| ------------------------------------------------- |
+| `specificationsTitle`                             |
+| `specificationsTableContainer`                    |
+| `specificationsTable`                             |
+| `specificationsTabsContainer`                     |
+| `specificationsTab`                               |
+| `specificationsTablePropertyHeading`              |
+| `specificationsTableSpecificationHeading`         |
+| `specificationItemProperty`                       |
+| `specificationItemProperty` + Specifications-name |
```
guerreirobeatriz, line 64:
```suggestion
| `specificationItemProperty` +  {specificationName} |
```

victorhmp:
Hi @BeatrizMiranda ! Just a quick update on this. We didn't forget about it! 😂
It's just that our releases are a bit slower this week since there is a big e-commerce event happening in Mexico and we're being especially careful with our deploys.

victorhmp:
Hi @BeatrizMiranda! I need `push` access to your fork so that I can rebase it with `master`.
I'm trying to release this along with another PR, and the CHANGELOG has conflicts (I guess because of the removed empty line in your branch).

luizgpriolli:
> Hi @BeatrizMiranda! I need `push` access to your fork so that I can rebase it with `master`.
> I'm trying to release this along with another PR, and the CHANGELOG has conflicts (I guess because of the removed empty line in your branch).

Access granted @victorhmp 
Please let me know if everything is ok

victorhmp:
> > Hi @BeatrizMiranda! I need `push` access to your fork so that I can rebase it with `master`.
> > I'm trying to release this along with another PR, and the CHANGELOG has conflicts (I guess because of the removed empty line in your branch).
> 
> Access granted @victorhmp
> Please let me know if everything is ok

Thanks @luizgpriolli ! Unfortunately, it didn't work :( I still can't push to the remote.

luizgpriolli:
> > > Hi @BeatrizMiranda! I need `push` access to your fork so that I can rebase it with `master`.
> > > I'm trying to release this along with another PR, and the CHANGELOG has conflicts (I guess because of the removed empty line in your branch).
> > 
> > 
> > Access granted @victorhmp
> > Please let me know if everything is ok
> 
> Thanks @luizgpriolli ! Unfortunately, it didn't work :( I still can't push to the remote.

Weird. I've send it again. I believe you must accept the invitation on your email first  @victorhmp.

victorhmp:
@luizgpriolli Thanks! That worked! 😄 .

victorhmp:
@BeatrizMiranda @luizgpriolli @klzns came up with a different solution for this.
What if we were to use the `specification.property` value as the value of a `data-specification` property in each of the `<td>`s?

That way, the HTML would look something like this:

Screen Shot 2020-06-02 at 14 16 39

And you would select these elements using CSS selectors such as: 

```css
.specificationItemProperty[data-specification='Filter Color:'] {
  width: 400px;
}
```

We think this would be a better solution because it makes more semantic sense. The specification name is much more the _data_ represented in the `td` tag than a CSS modifier. Since you would be able to achieve the same results in your customization, we think that might be a better solution.

What do you guys think?

BeatrizMiranda:
> @BeatrizMiranda @luizgpriolli
> 
> @klzns came up with a different solution for this.
> What if we were to use the `specification.property` value as the value of a `data-specification` property in each of the `<td>`s?
> 
> That way, the HTML would look something like this:
> 
> Screen Shot 2020-06-02 at 14 16 39
> 
> And you would select these elements using CSS selectors such as:
> 
> ```css
> .specificationItemProperty[data-specification='Filter Color:'] {
>   width: 400px;
> }
> ```
> 
> We think this would be a better solution because it makes more semantic sense. The specification name is much more the _data_ represented in the `td` tag than a CSS modifier. Since you would be able to achieve the same results in your customization, we think that might be a better solution.
> 
> What do you guys think?
> 
> Also, we'd have no problem in modifying the current implementation to achieve this, since we know that it would mean more work for you guys. But if you'd like to implement this new solution yourselves (if you guys think it's a good idea), you're welcome to 😄 .

I didn't know that vtex css allows us to use this haha
I made the change but how do I set this on the doc?

victorhmp:
> I didn't know that vtex css allows us to use this haha
> I made the change but how do I set this on the doc?

@BeatrizMiranda Yep, it does! 😄 
About the documentation, we don't have a clear solution for it yet, but we'll update it as soon as we do!

react/components/ProductSpecifications/index.js
```diff
@@ -1,4 +1,5 @@
 import React, { Fragment, useState } from 'react'
+import slugify from 'slugify'
```
victorhmp, line 2: No need for this import now 🙂

```suggestion
```

PR #780, opened 2020-06-02
Feature/specification name css

victorhmp:

This is a rebase of #777 .


victorhmp:
This is not necessary.

PR #786, opened 2020-06-05
contentType prop to the ProductImages Component

pedromtec:

I'm building the App Store PDP and I would like to filter the content type of ProductImages component.
The PDP uses the ProductImages block to render images and videos, but both cannot be displayed at the same time and the component doesn’t  allow hidden videos.

The prop `contentType` has three possible values:
`all` ( default ): Show videos and images  
`images`: Show images
`videos`: Show videos

[Workspace](https://pedro--extensions.myvtex.com/google-shopping/p)

#### Checklist/Reminders

- [ ] Updated `README.md`.
- [x] Updated `CHANGELOG.md`.

Screen Shot 2020-06-05 at 10 54 47 AM

#### Type of changes

✔️ | Type of Change
---|---
_ | Bug fix 
✔️| New feature 
_ | Breaking change 
_ | Technical improvements

### DIFF
--- a/react/components/ProductImages/Wrapper.js
+++ b/react/components/ProductImages/Wrapper.js
@@ -64,6 +64,7 @@ const ProductImagesWrapper = props => {
       contentOrder={contentOrder}
       // Deprecated
       zoomProps={props.zoomProps}
+      contentType={props.contentType}
     />
   )
 }
--- a/react/components/ProductImages/Wrapper.js
+++ b/react/components/ProductImages/Wrapper.js
@@ -62,9 +62,9 @@ const ProductImagesWrapper = props => {
       showNavigationArrows={showNavigationArrows}
       showPaginationDots={showPaginationDots}
       contentOrder={contentOrder}
+      contentType={props.contentType}
       // Deprecated
       zoomProps={props.zoomProps}
-      contentType={props.contentType}
     />
   )
 }
--- a/react/components/ProductImages/index.js
+++ b/react/components/ProductImages/index.js
@@ -29,6 +29,7 @@ const ProductImages = ({
   zoomFactor,
   // Deprecated
   zoomProps,
+  contentType = "all"
 }) => {
   if (hiddenImages && !Array.isArray(hiddenImages)) {
     hiddenImages = [hiddenImages]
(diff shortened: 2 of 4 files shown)


react/components/ProductImages/index.js
```diff
@@ -29,6 +29,7 @@ const ProductImages = ({
   zoomFactor,
   // Deprecated
   zoomProps,
+  contentType = 'all',
```
victorhmp, line 32: Minor detail, but I would move this **above** the `// Deprecated` comment

react/components/ProductImages/index.js
```diff
@@ -38,23 +39,31 @@ const ProductImages = ({
...
-      alt: image.imageText,
-      thumbUrl: image.thumbnailUrl || image.imageUrl,
-    }))
-  const videos = allVideos.map(video => ({
-    type: 'video',
-    src: video.videoUrl,
-    thumbWidth: 300,
-  }))
+  const images =
+    contentType !== 'videos'
```
victorhmp, line 43: Just to make the code easier to understand, you could store this check in a variable, such as `shouldIncludeImages` and then this conditional would look a bit clearer:

```suggestion
    shouldIncludeImages
```

react/components/ProductImages/index.js
```diff
@@ -38,23 +39,31 @@ const ProductImages = ({
...
+          .map(image => ({
+            type: 'image',
+            url: image.imageUrls ? image.imageUrls[0] : image.imageUrl,
+            alt: image.imageText,
+            thumbUrl: image.thumbnailUrl || image.imageUrl,
+          }))
+      : []
+
+  const videos =
+    contentType !== 'images'
```
victorhmp, line 59: The same idea applies here. 

```suggestion
    shouldIncludeVideos
```

victorhmp: Another advantage to abstracting this into a variable is that we could make the conditional more complex in the future (maybe when another prop is added), and this part of the code, and the understanding of it, would remain the same 😃 .

pedromtec: Yeah, it looks good to me too. Thanks @victorhmp .

react/components/ProductImages/Wrapper.js
```diff
@@ -64,6 +64,7 @@ const ProductImagesWrapper = props => {
       contentOrder={contentOrder}
       // Deprecated
       zoomProps={props.zoomProps}
+      contentType={props.contentType}
```
iaronaraujo, line 67: Could you please also move this to be above the comment?

PR #806, opened 2020-07-14
Add recursive input values in the right place

klzns:

Fix input values when it's inside an assembly option.

#### How to test it?

1. Open https://breno--storecomponents.myvtex.com/custom-bell/p
2. Click Customize in "Bells add-ons 1-3 lines"
3. Click "Add 1-3-Lines"
4. Fill the inputs "Line 1", "Line 2", and "Line 3".
5. Click "Done"
6. Click "Add To Cart"
7. Go to cart page
8. Once in the Cart page open dev tools and type in the console:
`vtexjs.checkout.orderForm.items[2].assemblies`

It should return an object with the values input.

Compare it with master workspace which doesn't have this fix and the last step won't work.

We should refactor the Product Customizer solution. It does not provide enough info for this solution to be future proof. It will break if two products have the same input values for instance.

#### Related to / Depends on

n/a

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+### Fixed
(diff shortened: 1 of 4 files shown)


Klynger:
Nice description

PR #814, opened 2020-07-16
Remove unneeded html parsing libs

kaisermann:
Another go at #810.

**The problem with the last solution:**

`html-react-parser` supports both `browser` and `node` environments by using the `html-dom-parser`, which has two different entry points: one for the `browser` and one for `node`. It does this by setting the `main` and `browser` properties in its `package.json`. However, the #810 changes were breaking our SSR because when we built the new version of the `store-components` app, the `react builder` would use the `browser` version of `html-dom-parser`, which uses DOM methods that are not available in `node`. 

**The previously implemented solution of this PR:**

~We can bypass the `html-react-parser`'s entry point and explicitly use the server/`node` version. This is far from ideal, but as you can see in the image below, it's way lighter than the current solution: it doesn't use `node`'s `buffer` module and there's no more duped `entities.json`.~

~Current solution on the left, new solution on the right~

image

~Instead of going from `46.25kb` to `6.34kb`, this PR makes the bundle go from `46.25kb` to `23.78` (51%).~

~**Notes:**~

- ~Since the server version of `html-dom-parser` uses `htmlparser2`, we would still have a dependency on `node`'s `events` module. You can see it on the far right side of the image above.~

**The current solution:**

We already have `insane`, a very lightweight HTML sanitizer. Let's use it and be done with that.

### DIFF
--- a/react/modules/sanitizedHTML.tsx
+++ b/react/modules/sanitizedHTML.tsx
@@ -35,10 +35,17 @@ function getDangerousSanitizedHTML(html: string, opts: SanitizeOpts = {}) {
   return { __html: sanitizeHTML(html, opts) }
 }
 
-function useSanitizedHTML(html: string, opts: SanitizeOpts = {}) {
+function useSanitizedHTML(
+  html: string,
+  { allowedAttributes, allowedClasses, allowedTags }: SanitizeOpts = {}
+) {
   const sanitizedHTML = useMemo(() => {
-    return getDangerousSanitizedHTML(html, opts)
-  }, [html, opts])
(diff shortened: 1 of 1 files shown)


react/package.json
```diff
@@ -13,6 +13,8 @@
     "graphql": "~0.13.2",
     "graphql-tag": "~2.8.0",
     "hoist-non-react-statics": "^3.3.0",
+    "html-dom-parser": "^0.3.0",
+    "html-react-parser": "^0.13.0",
```
Klynger, line 17: Is this still needed?

kaisermann: Nooope!

PR #823, opened 2020-07-21
Use originalName in SKU Selector to avoid CSS classes to change based on translation

klzns:

Add fields necessary to fix issue regarding fields being translated and changing the CSS classes applied.

#### How to test it?

Fixed:
https://breno--miriadeit.myvtex.com/?__bindingAddress=www.miriade.com/en
Class with `nero` applied:
image

Compare it with the current wrong behavior:
https://brenooff--miriadeit.myvtex.com/?__bindingAddress=www.miriade.com/en
Class with `black` was applied
image

none

#### Related to / Depends on

Depends on: https://github.com/vtex-apps/store-resources/pull/126

### DIFF
--- a/react/components/SKUSelector/components/Variation.tsx
+++ b/react/components/SKUSelector/components/Variation.tsx
@@ -60,7 +60,7 @@ const Variation: FC<Props> = ({
   sliderDisplayThreshold,
   sliderItemsPerPage,
 }) => {
-  const { name, options } = variation
+  const { originalName, name, options } = variation
 
   const visibleItemsWhenCollapsed = maxItems - ITEMS_VISIBLE_THRESHOLD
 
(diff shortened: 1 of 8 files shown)


react/components/SKUSelector/components/Variation.tsx
```diff
@@ -75,7 +75,7 @@ const Variation: FC<Props> = ({
     },
   } = useProduct()
 
-  const displayImage = isColor(name)
+  const displayImage = isColor(originalName)
```
jgfidelis, line 78: lol, can we delete the colors translation array now?

klzns: Nope, because we still depend on the language it was written in the Catalog.

PR #829, opened 2020-07-30
sku selector - use originalName to match with visibleVariations props

jgfidelis:

After we translated the names, I think people using the `visibleVariations` prop started to have problems.

https://fidelis--miriadeit.myvtex.com/sneakers-vally-new-mir/p?__bindingAddress=www.miriade.com/es

#### How to test it?

[Workspace](Link goes here!)

### DIFF
--- a/react/components/SKUSelector/Wrapper.tsx
+++ b/react/components/SKUSelector/Wrapper.tsx
@@ -176,8 +176,6 @@ const SKUSelectorWrapper: StorefrontFC<Props> = props => {
     props.visibleVariations
   )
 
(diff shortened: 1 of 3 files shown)


react/components/SKUSelector/Wrapper.tsx
```diff
@@ -176,6 +176,8 @@ const SKUSelectorWrapper: StorefrontFC<Props> = props => {
     props.visibleVariations
   )
 
+  console.log( 'teste VARIATIONS: ', {variations, skuSpecifications, visibleVariations: props.visibleVariations})
```
klzns, line 179: Remove

PR #831, opened 2020-08-03
Run updated eslint preset

kaisermann:

- Update tooling to latest version
- Run eslint/prettier on the project

Disclaimer: 

- Switched some `enum`s to `const enum`s.


react/components/SKUSelector/Wrapper.tsx
```diff
@@ -69,15 +76,21 @@ const getVariationsFromSpecifications = (
...
-  shouldNotShow: boolean,
+const useVariations = ({
+  skuItems,
+  skuSpecifications,
+  shouldNotShow,
+  visibleVariations,
+}: {
+  skuItems: ProductItem[]
+  skuSpecifications: SkuSpecification[]
+  shouldNotShow: boolean
```
Klynger, line 91: Just to confirm, these props can't be `undefined`, right?

react/components/SKUSelector/components/SKUSelector.tsx
```diff
@@ -71,21 +73,29 @@ const isSkuAvailable = compose<
...
-  variationCount: number,
+const showItemAsAvailable = ({
+  possibleItems,
+  selectedVariations,
+  variationCount,
+  isSelected,
+}: {
+  possibleItems: SelectorProductItem[]
+  selectedVariations: SelectedVariations
+  variationCount: number
```
Klynger, line 84: Same as the other comment

react/typings/storefront.d.ts
```diff
@@ -1,13 +1,16 @@
 import { FunctionComponent } from 'react'
 
+type GenericObject = Record<string, any>
```
Klynger, line 3: ✨

react/components/SKUSelector/Wrapper.tsx
kaisermann: Don't know, but even if they are `undefined` they will always be passed as `undefined` (lol)

PR #841, opened 2020-08-12
doc: fix ProductName showSku prop name

MatheusR42:
Fix doc

The prop name was wrong on the docs.


klzns:
@allcontributors add @MatheusR42 for docs

PR #850, opened 2020-08-25
CHK-177: Integrate shipping simulator with order-manager

lucasecdb:

This PR updates the `ShippingSimulator` component to be integrated with `order-manager`, i.e. it will first read the postal code value from the selected delivery address on the `orderForm` (if it can do so) and autofill the postal code input and prefetch the shipping simulation response with it. If the `orderForm` does not have a delivery address or we can't read it (if the user has been identified but not logged in, the shipping information will be masked and we won't be able to use it to simulate the shipping), we will simply show the input as it was before, but after the user has inputted their postal code we will save it in the `orderForm` so they don't have to input it again in other product pages.

It is important to notice that this new behavior will only work if the store is using `vtex.checkout>=1`.

#### How to test it?

[Store with `vtex.checkout@2`](https://lucas--checkoutio.myvtex.com/teste-so-delivery-copy-256--copy-257--copy-261--copy-262-/p).

[Store with `vtex.checkout@0`](https://lucas--storecomponents.myvtex.com/wood-clock/p).

ss

N/A

#### Related to / Depends on

N/A


klzns:
Why this only works with checkout 1?

lucasecdb:
> Why this only works with checkout 2?

@klzns I was under the impression I couldn't read the `orderForm` with Minicart v1, but then I realised that I could use the `OrderFormContext` in `store-resources` and have a similar behavior as with the `order-manager`, `order-shipping`, etc. I'll try to integrate it with store resources.

klzns:
Got it to work with the new minicart
https://shipping--storecomponents.myvtex.com/wood-clock/p

PR #854, opened 2020-09-03
Add props to UserAddress so that only city, state, and/or postal code can be displayed

polishq:

The US 1st Party Apps team is developing a solution for BOPIS (Pickup in Store) that utilizes the UserAddress component from this repo. Instead of using this component to display a delivery address, we wish to use it to display the user's location -- either their city and state, or their postal code (or both). Their location will affect the availability messaging they are shown on products shelves and the PDP. This PR allows the UserAddress component to display the desired information to the user.

#### How to test it?

The new version is linked here: https://geolocation--sandboxusdev.myvtex.com/
And has been configured to show city, state, postal code, and no "Order to" prefix.

geolocation--sandboxusdev myvtex com_

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
+### Added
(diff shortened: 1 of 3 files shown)


polishq:
Note: There does not seem to be documentation for the UserAddress component

klzns (approved):

PR #858, opened 2020-09-06
Dont render when not on screen

tlgimenes:
Currently, the BackToTopButton is rendered to the DOM and hidden by using CSS. 

I thinks this is a bad approach, since we spend precious "hydration" time computing a button that will be hidden using CSS. Also, this approach makes it harder to implement techniques like CriticalCSS.

To address this problem, I suggest only rendering the button once it's visible.

#### How to test it?

[Workspace](https://gimenes--carrefourbr.myvtex.com/?__debugCriticalCSS=true)


rafaelbac (approved):
LGTM

lbebber (approved):

PR #863, opened 2020-09-17
Updates to support new Shopper Location app

polishq:

A few updates to support the Shopper Location app and Buy Online / Pick Up In Store solution that the US 1st Party Apps team is developing.

**Shipping Simulator**
When the Order Form Optimization feature flag is enabled, the shipping simulator component is pre-filled with the postal code from the shopper's orderForm. This is good; however, just the act of rendering this component causes all of the orderForm's address fields other than the postal code to be emptied. This PR adds a prop that controls whether interacting with the shipping simulator modifies the address in the orderForm. 

**User Address**
This PR adds a new prop to this component that allows the component to be rendered even if no shipping address exists in the user's orderForm. If no address exists, the component will render an intl message like "Add Location". I've also tried to adjust the code to add support for the newer `vtex.modal-layout` in addition to keeping legacy support for `vtex.modal`. Finally, I've added an event listener for `locationUpdated` from `vtex.shopper-location` so that the orderForm can be refetched if the shopper changes their address.  

#### How to test it?

[Example workspace with UserAddress in header and ShippingSimulator on PDP](https://geolocation--sandboxusdev.myvtex.com/invicta-reserve-bolt-zeus-mens-automatic-53-mm-stainless-steel-case-silver-white-dial-model-27108/p)
(`vtex.shopper-location` will normally attempt to automatically determine the shopper's location. This feature has been disabled in this workspace in order to demonstrate the behavior when the shopper's address is empty.)

shopper-location

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,11 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
+### Added
+- `shouldUpdateOrderForm` prop to `ShippingSimulator`
+- `showIfEmpty` prop to `UserAddress` so that shopper with no address can add one
+- Event listener to `UserAddress` for `locationUpdated` event from `vtex.shopper-location`
+
 ## [3.130.0] - 2020-09-21
 ### Added
 - `product-sku-attributes` interface
--- a/react/components/UserAddress/index.js
+++ b/react/components/UserAddress/index.js
@@ -1,5 +1,5 @@
 import PropTypes from 'prop-types'
-import React from 'react'
+import React, { useEffect } from 'react'
 import { graphql } from 'react-apollo'
 import { useCssHandles } from 'vtex.css-handles'
 import ADDRESS_QUERY from 'vtex.store-resources/QueryAddress'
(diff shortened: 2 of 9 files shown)


react/components/UserAddress/index.js
```diff
@@ -16,12 +16,22 @@ const UserAddress = ({
...
 }) => {
   const { orderForm } = addressQueryProp
-  const { shippingData } = orderForm || {}
   const handles = useCssHandles(CSS_HANDLES)
 
-  if (!orderForm || !shippingData || !shippingData.address) {
+  useEffect(() => {
+    window.addEventListener('locationUpdated', () => {
+      addressQueryProp.refetch()
+    })
```
klzns, line 27: Missing a removeEventListener otherwise it will lead to a memory leak

klzns (changes requested):

react/components/UserAddress/index.js
```diff
@@ -16,12 +16,28 @@ const UserAddress = ({
...
+
+    return () => {
+      window.removeEventListener('locationUpdated', () => {
+        addressQueryProp.refetch()
```
klzns, line 31: This will not work.

It should be something like this:

```js
const handleLocationUpdated = () => addressQueryProp.refetch()

window.addEventListener('locationUpdated', handleLocationUpdated)

return () => {
  window.removeEventListener('locationUpdated', handleLocationUpdated)
}
```

It should reference the same function

klzns (changes requested):

klzns (approved):

guerreirobeatriz (changes requested):

CHANGELOG.md
```diff
@@ -7,6 +7,11 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
+### Added
+- `shouldUpdateOrderForm` prop to `ShippingSimulator`
+- `showIfEmpty` prop to `UserAddress` so that shopper with no address can add one
+- Event listener to `UserAddress` for `locationUpdated` event from `vtex.shopper-location`
```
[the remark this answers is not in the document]
polishq: @guerreirobeatriz There does not seem to be any existing documentation for the `UserAddress` component.

guerreirobeatriz (approved):

PR #865, opened 2020-09-18
Feature/product sku names

wender:
#### What problem is this solving?
Provide a list of the selected variations for the current product, also provide a source dependency to the `product-summary` 

#### How to test it?
Add the `product-sku-names` block to any block below `store.product` (Product template). For example:

```json
  "store.product": {
    "children": [
      "flex-layout.row#product",
    ]
  },
  "flex-layout.row#product": {
    "children": [
      "product-sku-names"
    ]
  },
```

[Check it working here](https://wender--alssports.myvtex.com/nixon-regulus-watch/p)

#### Screenshots or example usage:
image

PR #866, opened 2020-09-21
Use TS

klzns:
Change to use TS and simplify some code for #865


react/ProductSKUNames.tsx
```diff
@@ -0,0 +1,74 @@
...
+] as const
+
+const getVariationsFromSelectedItem = (
+  productContext: ReturnType<typeof useProduct>
+) => {
+  if (
+    !productContext ||
+    !productContext.product ||
+    !productContext.skuSelector ||
+    !productContext.selectedItem
```
victorhmp, line 19:
```suggestion
    !productContext?.product ||
    !productContext?.skuSelector ||
    !productContext?.selectedItem
```

victorhmp (approved):

react/ProductSKUNames.tsx
kaisermann: @victorhmp In this case, I don't think it makes much of a difference

react/ProductSKUNames.tsx
```diff
@@ -0,0 +1,74 @@
...
+  if (selectedImageVariationSKU) {
+    const selected = productContext.product.items.filter(item => {
+      return item.itemId === selectedImageVariationSKU
+    })
+
+    if (selected.length) {
+      const [item] = selected
+
+      return item.variations
+    }
```
kaisermann, line 35:
```suggestion
    const selected = productContext.product.items.find(item => {
      return item.itemId === selectedImageVariationSKU
    })

    if (selected) {
      return selected.variations
    }
```

kaisermann (approved):

wender (approved):

PR #881, opened 2020-11-17
InputSearch instead of Input for the AutocompleteInput

pmarignan:
First of all, it's about the AutcompleteInput component.
My solution solves the problem when you click on the 'ok' button then disappears the results block
Also, it allows to have a 'Search button' on the iOs virtual keyboard.

It's because my customers on iOs doesn't understand what happens. They search something, a block appears and shows a product list but when they click on the 'ok' button, the search input lost the focus and disappears the results block. So, my changes can solve 

that.

#### How to test it?
You can test it on https://drimtest2--drimjuguetes.myvtex.com/ 

[Workspace](Link goes here!)

PullRequestAutoCompleInput

BEFORE: 
   For example, you can reproduce the problem on https://www.loja.philips.pt/
AFTER:
   With my changes you can test it on https://drimtest2--drimjuguetes.myvtex.com/ 

#### How does this PR make you feel? [:link:](http://giphy.com/)
https://media.giphy.com/media/mgqefqwSbToPe/giphy.gif

### DIFF
--- a/react/__mocks__/vtex.styleguide.js
+++ b/react/__mocks__/vtex.styleguide.js
@@ -61,6 +61,27 @@ export const Input = forwardRef(function Input(
   )
 })
 
+export const InputSearch = forwardRef(function Input(
+  { label, error, errorMessage, isLoading, prefix, suffix, ...props },
+  ref
+) {
+  return (
+    <label>
+      {label}
+      {prefix}
+      <input
+        type="search"
+        data-isloading={isLoading}
+        data-error={error}
+        data-errormessage={errorMessage}
+        ref={ref}
+        {...props}
+      />
+      {suffix}
+    </label>
+  )
+})
+
 export const Button = jest.fn(
   ({ isLoading, variation, block, children, ...props }) => {
     return (
--- a/react/__mocks__/vtex.styleguide.js
+++ b/react/__mocks__/vtex.styleguide.js
@@ -61,7 +61,7 @@ export const Input = forwardRef(function Input(
   )
 })
 
-export const InputSearch = forwardRef(function Input(
+export const InputSearch = forwardRef(function InputSearch(
   { label, error, errorMessage, isLoading, prefix, suffix, ...props },
   ref
 ) {
--- a/react/components/SearchBar/components/AutocompleteInput.js
+++ b/react/components/SearchBar/components/AutocompleteInput.js
@@ -1,7 +1,7 @@
 import React, { useRef, useEffect } from 'react'
 import PropTypes from 'prop-types'
 import classNames from 'classnames'
-import { Input } from 'vtex.styleguide'
+import { InputSearch } from 'vtex.styleguide'
 import { ExtensionPoint, useChildBlock } from 'vtex.render-runtime'
 import { useCssHandles, applyModifiers } from 'vtex.css-handles'
 import { IconSearch, IconClose } from 'vtex.store-icons'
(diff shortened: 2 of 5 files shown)


react/__mocks__/vtex.styleguide.js
```diff
@@ -61,6 +61,27 @@ export const Input = forwardRef(function Input(
   )
 })
 
+export const InputSearch = forwardRef(function Input(
```
klzns, line 64:
```suggestion
export const InputSearch = forwardRef(function InputSearch(
```

react/components/SearchBar/components/AutocompleteInput.js
```diff
@@ -172,8 +172,20 @@ const AutocompleteInput = ({
...
     <div className={handles.autoCompleteOuterContainer}>
       <div className={classContainer}>
+      <InputSearch 
+          ref={inputRef}
+          size="large"
+          value={value}
+          {...restProps}          
+        />
+        {/*        
         <Input
```
klzns, line 189: I think we can use `<Input type="search" ...` instead of using InputSearch. Changing to `InputSearch` would break current CSS customisations.

react/components/SearchBar/components/AutocompleteInput.js
```diff
@@ -172,8 +172,20 @@ const AutocompleteInput = ({
   })
 
   return (
+    <form action="#" onSubmit={e => {
```
klzns, line 175: Do we need this form handler? Or is the input search that is fixing the issue?

react/components/SearchBar/components/AutocompleteInput.js
```diff
@@ -172,8 +172,20 @@ const AutocompleteInput = ({
...
+      e.preventDefault()
+      e.stopPropagation()
+      e.nativeEvent.stopImmediatePropagation()
```
klzns, line 178: Add comments to why these are being used.

klzns (changes requested):
Thanks for the pull request! I left some comments.

Also, can you add an entry to the CHANGELOG.md, please? We follow this standard: https://keepachangelog.com/en/1.0.0/

The GitHub Action bot pointed some lint issues as well.

react/__mocks__/vtex.styleguide.js
pmarignan: This change on vtex.styleguide.js doesn't affect anything so you can commit it.

react/components/SearchBar/components/AutocompleteInput.js
pmarignan: Ok I try it ant I'll do it something later.

react/components/SearchBar/components/AutocompleteInput.js
pmarignan: Absolutely! We need the form tag with action attribute + the input search too to have the search button on the iOs virtual keyboard.

react/components/SearchBar/components/AutocompleteInput.js
pmarignan: For example, if I search 'batman' when you click Enter on the keyboard, the URL changes with the searched word like this: /batman?_q=batman&map=ft and immediatly after execute the form action '#' and change the URL /#?_q=&map=ft.
So, we need to capture the event onSubmit to avoid to go to the form action "#"

klzns:
Thanks for the pull request! I had to do some changes, but I can't commit to your master branch so I created another PR #882. Next time you can create a branch when opening a pull request so we can push commits and contribute to it.

PR #900, opened 2020-12-04
Feature: ProductImages `displayMode`

juliomoreira:

- Adds `displayMode` to ProductImages component. When set as `inline` it will not render the images in a carousel (`default` mode). This gives more flexibility to use the component without the extra markup produced when rendering in a carousel.

#### How to test it?

[Workspace](https://juliomoreira--fashioneurope.myvtex.com/simple-and-beautiful-t-shirt/p)

juliomoreira--fashioneurope myvtex com_simple-and-beautiful-t-shirt_p_skuId=31


docs/ProductImages.md
```diff
@@ -53,7 +53,7 @@
...
 | `ModalZoom` | `block` | Opens a modal for product image zooming. This prop's value must match the name of the block responsible for triggering the modal containing the product image for zooming (e.g. `modal-layout` from [Modal layout](https://vtex.io/docs/components/all/vtex.modal-layout/) app). Notice that the `ModalZoom` prop will work only if the `zoomMode` prop is set as `open-modal`. To learn more, check out the [Advanced Configuration section](#Advanced-Configuration). | `undefined` |
 | `contentType` | `enum` | Controls the type of content that will be displayed in the block. Possible values are: `images`, `videos`, or `all`. | `all` |
-
+| `displayMode` | `enum` | Enables you to choose between `default` and `inline` modes. `default` mode displays the images in a carousel and is compatible with all props available. `inline` mode will display only images (videos not supported yet) and with no extra markup. Useful if you have a complex layout. Props compatible with this mode are: `maxHeight`, `hiddenImages`, `zoomFactor`, `aspectRatio`,`ModalZoomElement` and `zoomMode`. Possible values are: `default` or `inline`. | `default` |
```
guerreirobeatriz, line 56:
```suggestion
| `displayMode` | `enum` | Defines how the product media should be displayed. Possible values are `default` (displays the product images and videos in a carousel) and `inline` (displays only the product images inline, with no extra markup). *Caution*: The `inline` value does not display product videos and it is only compatible with the `maxHeight`, `hiddenImages`, `zoomFactor`, `aspectRatio`,`ModalZoomElement`, and `zoomMode` props. | `default` |

```

guerreirobeatriz (changes requested):

react/components/ProductImages/index.js
```diff
@@ -72,27 +79,43 @@ const ProductImages = ({
...
+          thumbnailMaxHeight={thumbnailMaxHeight}
+          showPaginationDots={showPaginationDots}
+          thumbnailAspectRatio={thumbnailAspectRatio}
+          showNavigationArrows={showNavigationArrows}
+          thumbnailsOrientation={thumbnailsOrientation}
+          displayThumbnailsArrows={displayThumbnailsArrows}
+          // Deprecated
+          zoomProps={zoomProps}
+        />
+      )}
```
kaisermann, line 118: Nit: can you replace this ternary with an `if(...) { return ... } return ...`?  It's a tad more readable

react/components/ProductImages/utils/enums.js
```diff
@@ -8,4 +8,9 @@ export const THUMBS_ORIENTATION = {
   HORIZONTAL: 'horizontal',
 }
 
+export const DISPLAY_MODE = {
+  DEFAULT: 'default',
```
kaisermann, line 12: What do you think of `carousel` instead of `default`?

kaisermann (changes requested):

react/components/ProductImages/utils/enums.js
```diff
@@ -8,4 +8,9 @@ export const THUMBS_ORIENTATION = {
...
+  INLINE: 'inline',
```
kaisermann, line 13: What do you think of `list`?

kaisermann (changes requested):

guerreirobeatriz (approved):

react/components/ProductImages/index.js
juliomoreira: @kaisermann it is definitely possible, but, since both `<ProductImages />` and `<Carousel />` share a parent HTML Element `<div className={`${productImagesContainerClass} ${handles.content} w-100`}>` it would became a bit more complex in order to avoid redundancy, we would have to create a wrapper component like `<ProductImagesContainer />` and pass props around, etc... Are you ok with that?

kaisermann: Can't we just store that className in a variable 😁 ?

juliomoreira: Sure!

juliomoreira (commented):
@kaisermann applied your suggestions

kaisermann:
@juliomoreira Thanks! We're going to bring this to our next grooming session and see if it's going to be merged as it is!

kaisermann (approved):
Just need a rebase first

PR #912, opened 2021-01-18
Feature/thumb vid class name

carafizi1:

I'm adding a aditional className to a thumbnail when it links to a video, such as following:
image

#### How to test it?

I've linked this feature in a workspace down below directly to a product that has both a video and a image. A video thumbnail should have a aditional thumbVid class on it.

thumbVid

[Workspace](https://carafizi--acctglobal.myvtex.com/luiz/p)

A simple "play" button added using CSS over the thumbnail so a user knows its links to a video.

image

#### How does this PR make you feel? [:link:](http://giphy.com/)

image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
+### Added
+- `thumbVid` cssHandles aditional class on `Product Image` to differ video thumbnails from image thumbnails
+
 ## [3.138.3] - 2021-01-13
 
 ### Fixed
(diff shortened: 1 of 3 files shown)


icazevedo (changes requested):
Code LGTM, but you should add the new CSS Handle to the docs.

carafizi1:
@icazevedo Just updated the docs!

carafizi1 (commented):
Just updated the docs!

icazevedo (approved):
LGTM

klzns (approved):
Thanks for the pull request. I changed the solution a bit to use a CSS Handle modifier instead of creating a new handle. I think that having both handles `thumbImg` and `thumbVideo` in the same element would be confusing.

igorbrasileiro (approved):

PR #917, opened 2021-01-27
Custom Collapse on product-description

khrizzcristian:
We can limit the display of the description depending on the number of characters on the desktop and mobile.

[Workspace](https://description--diverta.myvtex.com/adunarea-umbrelor-yne01878/p)
The app use the following code 
```
"product-description": {
    "props": {
        "customCollapse": true,
        "desktopLimitCharacters": 250,
        "mobileLimitCharacters": 150
    }
  }
```


react/ProductDescription.tsx
```diff
@@ -55,14 +66,31 @@ function ProductDescription(props: PropsWithChildren<Props>) {
...
+  let limitCharacters = 0;
+  if(customCollapse) {
+    desktopLimitCharacters = Number.isNaN(desktopLimitCharacters) ? defaultDesktopCharacters : desktopLimitCharacters;
+    mobileLimitCharacters = Number.isNaN(mobileLimitCharacters) ? defaultMobileCharacters : mobileLimitCharacters;
+    limitCharacters = isMobile ? mobileLimitCharacters : desktopLimitCharacters;
+  }
 
-  const { collapseContent = true, title } = props
+  const [ filteredDescription, setFilteredDescription ] = useState(description.substring(0, limitCharacters));
+  const [ showMoreButton, setShowMoreButton ] = useState(true)
```
icazevedo, line 88: You can't call hooks conditionally. You should move these calls up.

react/ProductDescription.tsx
```diff
@@ -55,14 +66,31 @@ function ProductDescription(props: PropsWithChildren<Props>) {
...
+
+  const changeDescription = () => {
+    setFilteredDescription(description)
+    setShowMoreButton(false)
+  }
```
icazevedo, line 93: Since their behavior is always going to be related, use a single state and make them react to it. Ex:

```ts
const [isCustomCollapsed, setCustomCollapsed] = useState(customCollapse)
```

then your `changeDescription` becomes setCustomCollapsed(false)

react/ProductDescription.tsx
```diff
@@ -55,14 +66,31 @@ function ProductDescription(props: PropsWithChildren<Props>) {
...
+  let {desktopLimitCharacters = defaultDesktopCharacters, mobileLimitCharacters = defaultMobileCharacters} = props
+  let limitCharacters = 0;
+  if(customCollapse) {
+    desktopLimitCharacters = Number.isNaN(desktopLimitCharacters) ? defaultDesktopCharacters : desktopLimitCharacters;
+    mobileLimitCharacters = Number.isNaN(mobileLimitCharacters) ? defaultMobileCharacters : mobileLimitCharacters;
+    limitCharacters = isMobile ? mobileLimitCharacters : desktopLimitCharacters;
+  }
 
-  const { collapseContent = true, title } = props
+  const [ filteredDescription, setFilteredDescription ] = useState(description.substring(0, limitCharacters));
```
icazevedo, line 87: Following the other suggestions I've made, you should make it react to `limitCharacters` and `isCustomCollapsed` (maybe use `useMemo`)

react/ProductDescription.tsx
```diff
@@ -55,14 +66,31 @@ function ProductDescription(props: PropsWithChildren<Props>) {
...
+  
+  const { collapseContent = true, title, customCollapse = false } = props
+
+  let {desktopLimitCharacters = defaultDesktopCharacters, mobileLimitCharacters = defaultMobileCharacters} = props
+  let limitCharacters = 0;
+  if(customCollapse) {
+    desktopLimitCharacters = Number.isNaN(desktopLimitCharacters) ? defaultDesktopCharacters : desktopLimitCharacters;
+    mobileLimitCharacters = Number.isNaN(mobileLimitCharacters) ? defaultMobileCharacters : mobileLimitCharacters;
+    limitCharacters = isMobile ? mobileLimitCharacters : desktopLimitCharacters;
+  }
```
icazevedo, line 85: This can all be inside an `useMemo`

icazevedo (changes requested):

react/ProductDescription.tsx
```diff
@@ -55,14 +66,31 @@ function ProductDescription(props: PropsWithChildren<Props>) {
...
   const { product } = useProduct()
+  const { isMobile } = useDevice()
 
   const description = props.description ?? product?.description
 
   if (!description) {
     return null
   }
+  
+  const { collapseContent = true, title, customCollapse = false } = props
```
icazevedo, line 77: Also, I think you're changing the default behavior here, since collapseContent used to default to false. That's a breaking change.

khrizzcristian: Yes. If you use customCollapse you will overwrite the default collapseContent behavior.

react/ProductDescription.tsx
```diff
@@ -55,14 +66,39 @@ function ProductDescription(props: PropsWithChildren<Props>) {
...
 
+  const { collapseContent = true, title, customCollapse = false } = props
   const description = props.description ?? product?.description
+  const [isCustomCollapsed, setCustomCollapsed] = useState(customCollapse)
+
+  const limitCharacters = useMemo(() => {
+    let {
+      desktopLimitCharacters = defaultDesktopCharacters,
+      mobileLimitCharacters = defaultMobileCharacters,
+    } = props
```
icazevedo, line 79: Move these variables out of the hook so they only execute when they change. Currently, if any prop change, it's going to run the memo again.

icazevedo (changes requested):
Just fix the comment I've left. Good job!

react/ProductDescription.tsx
icazevedo: Oh, I see! I didn't think `collapseContent` defaulted to `true`, my bad 😅

react/ProductDescription.tsx
```diff
@@ -55,14 +66,47 @@ function ProductDescription(props: PropsWithChildren<Props>) {
...
+  const { isMobile } = useDevice()
 
+  const { collapseContent = true, title, customCollapse = false } = props
   const description = props.description ?? product?.description
+  const [isCustomCollapsed, setCustomCollapsed] = useState(customCollapse)
+
+  const {
+    desktopLimitCharacters = defaultDesktopCharacters,
+    mobileLimitCharacters = defaultMobileCharacters,
+  }: any = props
```
icazevedo, line 78: Any particular reason you typed it as `any`? This should be avoided

react/ProductDescription.tsx
```diff
@@ -55,14 +66,47 @@ function ProductDescription(props: PropsWithChildren<Props>) {
...
+
+  const desktopLimitChars: any = useRef(desktopLimitCharacters)
+  const mobileLimitChars: any = useRef(mobileLimitCharacters)
```
icazevedo, line 81: I don't think you need that. You can just declare `const` variables inside the hook.

icazevedo (changes requested):

icazevedo (changes requested):
Oh, also, change the last commit's message to be more descriptive of the changes you've made

react/ProductDescription.tsx
khrizzcristian: err

I use that because of this warning. Let me know if I let the code as it is or I need to remove those lines.

react/ProductDescription.tsx
khrizzcristian: I'll remove that `any` . 
Please reply to the other message to know if I also remove the useRef and commit after.

react/ProductDescription.tsx
icazevedo: The problem your editor is pointing to is: you're trying to edit the outside variable inside the hook. Indeed that's not a good practice, but you don't actually need these updated values outside the hook, just the value you return (`limitCharacters`)

I'll leave a comment below so you can see a possible solution to this.

react/ProductDescription.tsx
```diff
@@ -55,14 +66,47 @@ function ProductDescription(props: PropsWithChildren<Props>) {
...
+  const limitCharacters = useMemo(() => {
+    if (isCustomCollapsed) {
+      desktopLimitChars.current = Number.isNaN(desktopLimitCharacters)
+        ? defaultDesktopCharacters
+        : desktopLimitCharacters
+      mobileLimitChars.current = Number.isNaN(mobileLimitCharacters)
+        ? defaultMobileCharacters
+        : mobileLimitCharacters
+
+      return isMobile ? mobileLimitChars.current : desktopLimitChars.current
```
icazevedo, line 92:
```suggestion
      const validDesktopLimitCharacters = Number.isNaN(desktopLimitCharacters)
        ? defaultDesktopCharacters
        : desktopLimitCharacters
      const validMobileLimitCharacters = Number.isNaN(mobileLimitCharacters)
        ? defaultMobileCharacters
        : mobileLimitCharacters

      return isMobile ? validMobileLimitCharacters : validDesktopLimitCharacters

**Review by icazevedo — APPROVED** [review]
LGTM

**Review by igorbrasileiro — APPROVED** [review]
LGTM

**Comment by igorbrasileiro** [comment]
Can you rewrite the message of this commit `qf`?

File: docs/ProductDescription.md [hunk 10]
```diff
@@ -39,6 +39,9 @@ This Component can be imported and used by any VTEX App.
 | `collapseContent` | `Boolean` | If true, whenever the product description is too big, it will collapse and show a "Show More" button. When false, it will never collapse and will always show the whole description. | `true` |
 | `title`           | `string`  | Defines a custom title for the description section. | `undefined` |
 | `classes` | `CustomCSSClasses` | Used to override default CSS handles. To better understand how this prop works, we recommend reading about it [here](https://github.com/vtex-apps/css-handles#usecustomclasses). Note that this is only useful if you're using this block as a React component. | `undefined` |
+| `customCollapse`           | `boolean` | Overrides the `collapseContent` prop and allows you to collapse content by number of characters. | `false` |
+| `desktopLimitCharacters`   | `number`  | Works only if `collapseContent` is `true`. It determine the number of characters shown for desktop devices. | `1000` |
+| `mobileLimitCharacters`    | `number`  | Works only if `collapseContent` is `true`. It determine the number of characters shown for mobile devices. | `500` |
```
klzns, line 44: I think it's better to create a prop called `ellipsisContent`. This prop would be a [responsive-value prop](https://github.com/vtex-apps/responsive-values).

So the user could do something like:

```tsx
"ellipsisContent": {
  "mobile": 500,
  "desktop": 1000
}
  ```

Or if it's the same value for both:

```tsx
"ellipsisContent": 600
```

If `ellipsisContent` is set, we can consider it's active.

messages/context.json
```diff
@@ -188,5 +188,6 @@
   "admin/editor.product-description.title-prop.title": "Title",
   "admin/editor.search-bar.title": "Search bar",
   "admin/editor.search-bar.placeholder.title": "Placeholder",
-  "store/back-to-top.label": "Back to top"
+  "store/back-to-top.label": "Back to top",
+  "store/customCollapse.readMore": "...read more"
```
klzns, line 192:
```suggestion
  "store/customCollapse.readMore": "…read more"
```

messages/en.json
```diff
@@ -192,5 +192,6 @@
   "admin/editor.toTop.title": "Back to top button",
   "admin/editor.toTop.description": "When clicked it takes the page to the top",
   "admin/editor.toTop.topPixel.title": "Number of pixels",
-  "admin/editor.toTop.topPixel.description": "When exceeding the defined pixel value, the button to return to the top will appear"
+  "admin/editor.toTop.topPixel.description": "When exceeding the defined pixel value, the button to return to the top will appear",
+  "store/customCollapse.readMore": "...read more"
```
klzns, line 196:
```suggestion
  "store/customCollapse.readMore": "…read more"
```

messages/es.json
```diff
@@ -192,5 +192,6 @@
   "admin/editor.toTop.title": "Back to top button",
   "admin/editor.toTop.description": "Al hacer clic, lleva la página a la parte superior",
   "admin/editor.toTop.topPixel.title": "Número de píxeles",
-  "admin/editor.toTop.topPixel.description": "Al exceder el valor de píxel definido, aparecerá el botón para volver a la parte superior."
+  "admin/editor.toTop.topPixel.description": "Al exceder el valor de píxel definido, aparecerá el botón para volver a la parte superior.",
+  "store/customCollapse.readMore": "...lee mas"
```
klzns, line 196:
```suggestion
  "store/customCollapse.readMore": "…lee mas"
```

messages/pt.json
```diff
@@ -192,5 +192,6 @@
   "admin/editor.toTop.title": "Botão Voltar ao topo",
   "admin/editor.toTop.description": "Quando clicado, leva a página ao topo",
   "admin/editor.toTop.topPixel.title": "Número de pixels",
-  "admin/editor.toTop.topPixel.description": "Ao exceder o valor de pixel definido, o botão para retornar ao topo aparecerá"
+  "admin/editor.toTop.topPixel.description": "Ao exceder o valor de pixel definido, o botão para retornar ao topo aparecerá",
+  "store/customCollapse.readMore": "...consulte mais informação"
```
klzns, line 196:
```suggestion
  "store/customCollapse.readMore": "…ler mais"
```

react/ProductDescription.tsx
```diff
@@ -1,20 +1,28 @@
...
 import { SanitizedHTML, DEFAULTS } from './components/SanitizedHTML'
 import GradientCollapse from './components/GradientCollapse'
 
+import './ProductDescription.css'
+
 const CSS_HANDLES = [
   'productDescriptionContainer',
   'productDescriptionTitle',
   'productDescriptionText',
+  'showMoreButtonSpan',
```
klzns, line 20:
```suggestion
  'readMoreButton',
```

react/ProductDescription.css
```diff
@@ -0,0 +1,13 @@
+.showMoreButtonSpan :global(.vtex-button):hover {
+  background-color: transparent;
+}
+
+.showMoreButtonSpan :global(.vtex-button):focus {
+  outline: none;
+}
+
+.showMoreButtonSpan :global(.vtex-button__label) {
```
klzns, line 9:
```suggestion
.readMoreButton :global(.vtex-button):hover {
  background-color: transparent;
}

.readMoreButton :global(.vtex-button):focus {
  outline: none;
}

.readMoreButton :global(.vtex-button__label) {
```

docs/ProductDescription.md
```diff
@@ -49,3 +52,4 @@ In order to apply CSS customizations in this and other blocks, follow the instru
 | `productDescriptionContainer` |
 | `productDescriptionTitle`     |
 | `productDescriptionText`      |
+| `showMoreButtonSpan`          |
```
klzns, line 55:
```suggestion
| `readMoreButton`              |
```

klzns (changes requested):
Thanks for the pull request!

I proposed a different API that would be simpler to manage it.

icazevedo (approved):

docs/ProductDescription.md
igorbrasileiro: I agree with that. What do you think about @khrizzcristian ?

igorbrasileiro (changes requested):
Only fix the order of properties, to that comment have a meaning

victorhmp:
Closing this since it looks to be stale.

PR #920, opened 2021-02-12
Fix searches with slash

thalytafabrine:
now, the problem of searching with slash will be solved directly in the rewriter, so that it also works with facet selection and not just full text. 
Therefore, we no longer need to put this encoded value for the slash here, it's only necessary to make the standard encode so the rewriter can differentiate the slash from the query in the url.

#### How to test it?

[Before](https://biscoind.myvtex.com/3-48m$2fs?_q=3-48m$2Fs&map=ft)
[After](https://encodeurl--biscoind.myvtex.com/3-48m%2fs?_q=3-48m/s&map=ft)
- note the value encoded in the URL

#### Related to / Depends on

Depends on: https://github.com/vtex/rewriter/pull/186

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+### Fixed
+- Searches with slash.
 
 ## [3.141.1] - 2021-02-17
 ### Fixed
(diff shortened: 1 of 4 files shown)


hiagolcm (commented):
I think the `react/utils/decodeForwardSlash.ts ` is not necessary anymore, right?

thalytafabrine:
> I think the `react/utils/decodeForwardSlash.ts ` is not necessary anymore, right?

@hiagolcm yes, this file was deleted 🙂

hiagolcm:
> > I think the `react/utils/decodeForwardSlash.ts ` is not necessary anymore, right?
> 
> @hiagolcm yes, this file was deleted 🙂

Oh of course. I completely missed the change

hiagolcm (approved):

PR #924, opened 2021-02-26
Push 'newsletterInput' pixel event after newsletter mount

igorbrasileiro:
Pixel listrak try to find newsletter input, but the element isn't rendered when the listrak script is loaded, so the pixel doesn't work properly.

#### How to test it?
[Workspace](https://brasileirovtex--eriksbikeshop.myvtex.com/)

1. Open the store page
2. Go to newsletter field
3. add an email and submit
4. check if a request is made to https://sca1.listrakbi.com/ 
image

Change render-runtime, but seems overkill.


react/components/Newsletter/index.js
```diff
@@ -36,6 +37,7 @@ class Newsletter extends Component {
 
   componentDidMount() {
     this.mounted = true
+    this.props.push({ event: 'pageView' })
```
victorhmp, line 40: Why did you choose this event? This could be an entirely new event, like `newsletter-form-read` for example.

Sending a `pageView` might generate duplicate events, right?

victorhmp (changes requested):

react/components/Newsletter/index.js
igorbrasileiro: @victorhmp, I was talking with @polishq because I don't have too much context about listrak and he showed me where the event should be captured, it's [here](https://github.com/vtex-apps/listrak-pixel/blob/master/react/index.tsx#L52). Listrak pixel already handle `vtex:pageView` event.

victorhmp: Ookay, that's fine then.

victorhmp (approved):

react/components/Newsletter/index.js
klzns: I think it's better to just add a new switch case in listrak-pixel app to listen to this new event, otherwise we would trigger duplicate pageViews for a single page which is not desirable.

klzns (changes requested):

klzns (approved):

PR #927, opened 2021-03-25
Docs: updating image docs

FlavioOdas:

Updating `Image` docs based on this [store-image PR](https://github.com/vtex-apps/store-image/pull/35)

### DIFF
--- a/docs/Image.md
+++ b/docs/Image.md
@@ -31,6 +31,8 @@ The `Image` is a VTEX block that allows to **add any image** in the store.
 | ------------- | ---------- | -------------------------------------------------------------------------- | - |
 | `src`         | `string!`  | Specifies the URL of an image                                              | - |
 | `alt`         | `string`   | Specifies an alternate text for an image                                   | - |
+| `inlineWidth` | `number \| string` | Specifies the inline width attribute of an image                   | - |
+| `inlineHeight`| `number \| string` | Specifies the inline height attribute of an image                  | - |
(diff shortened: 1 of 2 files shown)


docs/Image.md
```diff
@@ -31,6 +31,8 @@ The `Image` is a VTEX block that allows to **add any image** in the store.
 | ------------- | ---------- | -------------------------------------------------------------------------- | - |
 | `src`         | `string!`  | Specifies the URL of an image                                              | - |
 | `alt`         | `string`   | Specifies an alternate text for an image                                   | - |
+| `inlineWidth` | `number \| string` | Specifies the inline width attribute of an image                   | - |
+| `inlineHeight`| `number \| string` | Specifies the inline height attribute of an image                  | - |
```
guerreirobeatriz, line 35:
```suggestion
| `inlineWidth` | `number` / `string` | Specifies the image's inline width attribute  | - |
| `inlineHeight`| `number` / `string` | Specifies the image's inline height attribute | - |
```

guerreirobeatriz: I don't understand the scenarios in which the height/width attribute should be written as a string. Could you give me an example?

guerreirobeatriz (changes requested):

docs/Image.md
FlavioOdas: @guerreirobeatriz These attributes also accept `percentages` and `auto` values, so the string option will provide some coverage for those situations

image

PR #944, opened 2021-05-17
Add new prop to Shipping Simulator

LucasCastroJussi:

Give a freight visibility option.
If the product has gifts attached, you can choose a property where you choose to show the shipping options for each item, or add up all values ​​by modality.

#### How to test it?

1- Enter the following workspace: 
[Workspace](https://sdtio763--diageo.myvtex.com/whisky-johnnie-walker-red-label-embalagem-comemorativa-de--200-anos--1l-759895_pai/p)
2 - Add the sumAllValues ​​property with the value "true" in the component;
3 - See that now, the freight values ​​will be shown following the respective grouping of their types;

Before and with "sumAllValues" as false:

PR-1

PR-2

After with "sumAllValues" as true:

PR-3

PR-4


docs/ShippingSimulator.md
```diff
@@ -40,6 +40,7 @@ The Shipping Simulator block **estimates the shipping fee** based on a zip code
 | ----------------------- | --------- | --------------------------------------------------------------------------------------------- | ------------- |
 | `skuId`                 | `String`  | ID of the current product SKU                                                                 | -             |
 | `seller`                | `String`  | ID of the product seller                                                                      | -             |
+| `sumAllValues`          | `Boolean` | If the product has gifts or attachments, you can choose whether the shipping information will be grouped by shipping type or showing the shipping prices for each of the items individually. | `false`       |
```
victorhmp, line 43: Instead of creating a boolean prop, how about creating an enum?

We tend to stay away from boolean props as much as possible so that our component's APIs don't grow indefinitely.

For this use case, how about a `pricingMode` prop, with possible values: `individualItems` and `grouped`? Something along those lines.

Then in the code, `individualItems` would be the default case and `grouped` would be the behavior you've just implemented.

victorhmp (changes requested):
Thanks for the PR @LucasCastroJussi!

docs/ShippingSimulator.md
LucasCastroJussi: @victorhmp,

Thanks for the sugestion.

I will implement these changes and will soon return.

LucasCastroJussi: @victorhmp 

Corrections made.

Could you review it again, please?

react/components/ShippingSimulator/components/ShippingTable.js
```diff
@@ -45,14 +59,24 @@ const ShippingTable = ({ shipping }) => {
...
       <tbody className={styles.shippingTableBody}>
-        {slaList.map(shippingItem => (
-          <ShippingTableRow
-            key={shippingItem.id}
-            name={shippingItem.friendlyName}
-            shippingEstimate={shippingItem.shippingEstimate}
-            price={shippingItem.price}
-          />
-        ))}
+        {pricingMode === GROUPED ?
```
icazevedo, line 62: I think you can simplify this by just switching between the lists depending on the value of `pricingMode` and use just one map function. Something as:

```suggestion
        {(pricingMode === GROUPED ? slaSumValuesList : slaList).map(shippingItem
```

icazevedo (changes requested):
I've left a suggestion for you. Also, please fix the lint errors.

react/components/ShippingSimulator/components/ShippingTable.js
LucasCastroJussi: @icazevedo 

Corrections made.

Could you review it again, please?

victorhmp (approved):
LGTM!

victorhmp (changes requested):
Sorry, just noticed that the tests are falling :(
Could you fix them before we can merge this?

LucasCastroJussi:
@victorhmp I tried to correct.

Can you review again, please?

icazevedo (approved):

victorhmp (approved):

icazevedo:
@allcontributors add @LucasCastroJussi for code

PR #946, opened 2021-05-24
Changing used handle by product sku

brener-leite:

The "skuName" exported by "productName" app was using wrong CSS handle (it was using "handles.productBrand", but it should be "handles.productSku"), so I've just changed it.

### DIFF
--- a/react/ProductName.tsx
+++ b/react/ProductName.tsx
@@ -110,7 +110,7 @@ function ProductName({
         {name} {showBrandName && brandName && `- ${brandName}`}
       </span>
       {showSku && skuName && (
-        <span className={`${handles.productBrand} ${skuNameClass ?? ''}`}>
+        <span className={`${handles.productSku} ${skuNameClass ?? ''}`}>
           {skuName}
         </span>
       )}


victorhmp:
Hi @brener-leite !
Thanks for the PR! But could you please follow the PR template? It makes it much easier for us to review this and make sure it works. 
Could you provide a workspace where we can test this change? Or even better, a screenshot showing that it works? Also, could you please update the CHANGELOG file?

icazevedo (changes requested):
Thanks for the PR @brener-leite !
I want to hear @victorhmp opinion on this one. It's broken like that for 6 months, which means that some clients may be counting on this behavior. Do we risk fixing it or will we be forced to include both handles?

For me, none of the solutions is good. One means we're not offering no way to customize only this specific span, and the other forces the user to deal with overriding customization if `productBrand` has any override already.

Also @brener-leite , could you please add a changelog entry for the change?

igorbrasileiro:
@icazevedo I understood your point of view, but if some clients are using it that way, they are using the API wrongly. What we can do to mitigate the impact is use the release notes to comunicate about this fix.

victorhmp:
I'm closing this PR since it has been stale since May 24th.

PR #951, opened 2021-06-03
✨Added displayMode prop

anto90fg:

For SEO reason, there may be a need to have the link to the current product associated with the product name. For this reason, I have added the possibility of using a settable prop to include the product name in a link to the product page.

#### How to test it?

I insert a direct link to the plp, in which the component has been set and the prop set to true.

[Workspace](https://productname--itwhirlpool.myvtex.com/prodotti/lavaggio-e-asciugatura/lavatrici)

#### Screenshots

If the prop is set to true, then the behaviour is as shown in the screenshot below. 
image

If not set, or set to false, then the behaviour is the standard one.
image

##### Version
Version: 3.143.1
Proposed version: 3.144.0

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+
+## [3.143.1] - 2021-04-28
 ### Fixed
 - Makes main SearchBar placeholder translatable
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ## [Unreleased]
 
 ### Added
-- `showProductLink` prop to show the link associated with the product name.
+- `displayMode` prop useful for displaying the product name in plain text or embedded in the link to the product page.
 
 ## [3.144.0] - 2021-05-31
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,19 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ### Added
 - `displayMode` prop useful for displaying the product name in plain text or embedded in the link to the product page.
 
+## [3.147.0] - 2021-06-29
+
+### Added
+- `disableBlurAndTouchEndHandler` prop to the `SearchBar` component.
+
+## [3.146.0] - 2021-06-28
+### Added
+- `unavailable` modifier to `valueWrapper` CSS handle.
+
+## [3.145.0] - 2021-06-28
+### Added
+- `showTitle` prop to `ProductDescription` component.
+
 ## [3.144.0] - 2021-05-31
 
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,8 +6,15 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+
 ### Added
 - `showProductLink` prop to show the link associated with the product name.
+
+## [3.144.0] - 2021-05-31
+
+### Added
+- Add new prop to group Shipping-Simulator similar values
+
 ## [3.143.1] - 2021-04-28
 ### Fixed
 - Makes main SearchBar placeholder translatable
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,8 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
-
+### Added
+- `showProductLink` prop to show the link associated with the product name.
 ## [3.143.1] - 2021-04-28
 ### Fixed
 - Makes main SearchBar placeholder translatable
--- a/docs/ProductName.md
+++ b/docs/ProductName.md
@@ -43,6 +43,7 @@ The `ProductName` is a block responsible for **displaying the product name** alo
 | `showProductReference` | `Boolean` | Show product reference | `false`| 
 | `showBrandName` | `Boolean` | Show brand name | `false`| 
 | `classes` | `CustomCSSClasses` | Used to override default CSS handles. To better understand how this prop works, we recommend reading about it [here](https://github.com/vtex-apps/css-handles#usecustomclasses). Note that this is only useful if you're using this block as a React component. | `undefined` |
+| `showProductLink` | `Boolean` | Show the product link associated with the its name. | `false`| 
 
 ## Customization
 
--- a/docs/ProductName.md
+++ b/docs/ProductName.md
@@ -43,7 +43,7 @@ The `ProductName` is a block responsible for **displaying the product name** alo
 | `showProductReference` | `Boolean` | Show product reference | `false`| 
 | `showBrandName` | `Boolean` | Show brand name | `false`| 
 | `classes` | `CustomCSSClasses` | Used to override default CSS handles. To better understand how this prop works, we recommend reading about it [here](https://github.com/vtex-apps/css-handles#usecustomclasses). Note that this is only useful if you're using this block as a React component. | `undefined` |
-| `showProductLink` | `Boolean` | Show the product link associated with the its name. | `false`| 
+| `displayMode` | `enum` | Possible values: `linkToProductPage` (show the product link associated with its name) or `plainText` (to show only the product name). | `plainText`| 
 
 ## Customization
 
--- a/messages/context.json
+++ b/messages/context.json
@@ -8,6 +8,8 @@
   "admin/editor.productName.showSku.title": "admin/editor.productName.showSku.title",
   "admin/editor.productName.showProductReference.title": "admin/editor.productName.showProductReference.title",
   "admin/editor.productName.displayMode.title": "admin/editor.productName.displayMode.title",
+  "admin/editor.productName.displayMode.plainText": "admin/editor.productName.displayMode.plainText",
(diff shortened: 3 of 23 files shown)


docs/ProductName.md
```diff
@@ -43,6 +43,7 @@ The `ProductName` is a block responsible for **displaying the product name** alo
 | `showProductReference` | `Boolean` | Show product reference | `false`| 
 | `showBrandName` | `Boolean` | Show brand name | `false`| 
 | `classes` | `CustomCSSClasses` | Used to override default CSS handles. To better understand how this prop works, we recommend reading about it [here](https://github.com/vtex-apps/css-handles#usecustomclasses). Note that this is only useful if you're using this block as a React component. | `undefined` |
+| `showProductLink` | `Boolean` | Show the product link associated with the its name. | `false`|
```
victorhmp, line 46: Hi @anto90fg !
To keep our component's APIs as flexible as possible, without constantly making them bigger, we tend to stay away from boolean props.

In this case, to implement the functionality you want, how about creating a prop called `displayMode` (or something along those lines), which would accept the values `plain-text` and `link-to-product-page`? This way, if we need to add support to a new variation of this in the future, we just need to make the same prop accept one more value, instead of creating a whole new prop.

I'm not sold on these names I;'ve suggested 😅  Maybe @icazevedo has a better idea for them

victorhmp (changes requested):

docs/ProductName.md
icazevedo: I like the names, but I think they should be camel-cased. `plainText` and `linkToProductPage`.

react/ProductName.tsx
```diff
@@ -102,6 +109,37 @@ function ProductName({
     )
   }
 
+  if (showProductLink && productLink) {
+    return (
+      <Wrapper
+        className={`${handles.productNameContainer} mv0 ${className ?? ''}`}
+      >
+        <a
```
icazevedo, line 117: I don't think this is the better way of doing it. We usually use `render-runtime`'s `Link` component to redirect users to the product page. Take a look at this [example](https://github.com/vtex-apps/product-summary/blob/46ade16bc8715835b39281cacec32f4823179b87/react/ProductSummaryCustom.tsx#L140).

I also think we shouldn't duplicate all the code in case there's a link wrap or not. You should create something like:

```suggestion
        const LinkWrapper = showProductLink ? Link : Fragment
        
        // And then use it where you put the a tag previously
        <LinkWrapper {...linkProps}>
          // Rest of the code
        </LinkWrapper>
```

icazevedo (changes requested):
Thanks for the PR! I've left some required changes below.

react/ProductName.tsx
anto90fg: I fixed it, following the requested changes. You can find all changes in the "required changes" commit.

docs/ProductName.md
anto90fg: I fixed it, following the requested changes. You can find all changes in the "required changes" commit.

react/ProductName.tsx
```diff
@@ -138,6 +164,8 @@ function ProductNameWrapper(props: Props) {
   }
 
   const { product, selectedItem } = valuesFromContext
+  const displayMode =
+    props.displayMode !== undefined ? props.displayMode : 'plainText'
```
victorhmp, line 168: You don't need this check here, you can just set a default value in the actual `ProductName` component.

I prefer having it there to do as little work as possible inside this `ProductNameWrapper` component. It makes this whole file and components easier to understand. Also, since you're just setting a default value by checking if something is `undefined`, you can just use JS in your favor and not have to do this check manually :)

react/ProductName.tsx
```diff
@@ -102,27 +113,42 @@ function ProductName({
     )
   }
 
+  const LinkWrapper =
```
victorhmp, line 116: Switching between rendering a <Link/> or a `<Fragment/>` is a nice idea! But passing `className` and all of the necessary props for the `Link` even if we're rendering a `<Fragment />` is not a good side effect. 

And I'm pretty sure would also flood the browser's cursor with errors/warnings from React, since `Fragment`s should only receive either a `key` property or none.

Swapping out the `Fragment` for a `div` might be a partial solution, but it isn't great either, since a bunch of props would still be passed to it.

I suggest creating a simple functional component, in this same file, that simply decides what to render in the end, something like:

```js
const LinkWrapper = ({ displayMode, linkProps, className, children }) => {
  if (displayMode === 'plainText') {
    return (
      <Fragment>
        {children}
      </Fragment>
    )
  }

  return (
    <Link className={className} {..linkProps}>
     {children}
    <Link />
  )
}
```

react/ProductName.tsx
```diff
@@ -76,6 +84,9 @@ function ProductName({
   skuName,
   brandName,
   productReference,
+  displayMode,
```
victorhmp, line 87: Add a default value for this prop right here

victorhmp (changes requested):

react/ProductName.tsx
anto90fg: Done!

react/ProductName.tsx
anto90fg: I completely agree with the creation of a functional component!

react/ProductName.tsx
```diff
@@ -59,6 +61,12 @@ type Props = {
   tag?: 'div' | 'h1' | 'h2' | 'h3'
   /** Used to override default CSS handles */
   classes?: CssHandlesTypes.CustomClasses<typeof CSS_HANDLES>
+  /** Show product link with the product name */
+  displayMode?: 'plainText' | 'linkToProductPage'
+  /** Product link */
+  productLink?: string
+  /** Product link */
```
icazevedo, line 68:
```suggestion
  /** Product id */
```

docs/ProductName.md
icazevedo: Please update the docs to reflect these changes 🙏

CHANGELOG.md
```diff
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
+### Added
+- `showProductLink` prop to show the link associated with the product name.
```
icazevedo, line 11: Fix the prop name here too

icazevedo (changes requested):

CHANGELOG.md
anto90fg: Fixed :)

react/ProductName.tsx
anto90fg: Fixed :)

messages/context.json
```diff
@@ -7,6 +7,7 @@
   "admin/editor.productName.showBrandName.title": "admin/editor.productName.showBrandName.title",
   "admin/editor.productName.showSku.title": "admin/editor.productName.showSku.title",
   "admin/editor.productName.showProductReference.title": "admin/editor.productName.showProductReference.title",
+  "admin/editor.productName.showProductLink.title": "Showing the link associated with the product name",
```
camilavcoutinho, line 10:
```suggestion
  "admin/editor.productName.showProductLink.title": "admin/editor.productName.showProductLink.title",
```

camilavcoutinho (changes requested):
For EN/PT/ES all good, the only thing that I think needs changes is context.json.

react/ProductName.tsx
```diff
@@ -102,31 +113,60 @@ function ProductName({
...
+        )}
+      </LinkWrapper>
     </Wrapper>
   )
 }
 
+/**
+ * Shows the link associated with the product name or plain text depending on the displayMode.
+ */
+const LinkWrapper = ({ displayMode, linkProps, className, children }: any) => {
```
igorbrasileiro, line 158: Can you type LinkWrapper props instead of any? Would be nice for readability and typecheck

react/ProductName.tsx
```diff
@@ -102,31 +113,60 @@ function ProductName({
...
+  if (displayMode === 'plainText') {
+    return <Fragment>{children}</Fragment>
+  }
```
igorbrasileiro, line 161: If you type props properly, you don't need to use this Fragment component, probably.

igorbrasileiro (changes requested):

react/ProductName.tsx
anto90fg: Done! :)

messages/context.json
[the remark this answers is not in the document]
anto90fg: Done :)

react/ProductName.tsx
anto90fg: How I could avoid using the fragment? I think it's a good choice to use it. But if you have a suggestion on how to remove it, I welcome it. :)

icazevedo: I think you could just `return children` but I also think this is a matter of style. I see no problem with it staying as it is

icazevedo (approved):
LGTM

victorhmp:
Hi @anto90fg !
I tried to review this again and it looks like your workspace is not working anymore :(
The one thing I was trying to check was how this shows up in the Site Editor. I'm pretty sure that the values you set in `enumNames` (used in this component's schema) is how each option shows up in the Site Editor. And as it stands, it would result in a bad UX for Site Editor users. Please use a translatable message for the `enumNames`, so that they can be more intuitive and also for i18n support.

anto90fg:
> Hi @anto90fg !
> I tried to review this again and it looks like your workspace is not working anymore :(
> The one thing I was trying to check was how this shows up in the Site Editor. I'm pretty sure that the values you set in `enumNames` (used in this component's schema) is how each option shows up in the Site Editor. And as it stands, it would result in a bad UX for Site Editor users. Please use a translatable message for the `enumNames`, so that they can be more intuitive and also for i18n support.

Hi @victorhmp , I inserted the enum name messages and moved the LinkWrapper component above its call, since the previous definition generated a lint error (the component was called before its definition).
I also updated the workspace ;)

igorbrasileiro:
The code LGTM

igorbrasileiro (approved):

victorhmp (approved):

camilavcoutinho (approved):

igorbrasileiro:
@anto90fg can you resolve the conflicts?

anto90fg:
> @anto90fg can you resolve the conflicts?

Done!

PR #953, opened 2021-06-07
Disable select options in dropdown

razvanudream:

Similar to isImpossible prop from SelectorItem component where 'o-20' (opacity) is applied for options when displayMode is default, we should have the ability to disable select options (from dropdown) as well, in the case of displayMode: select

I know hideImpossibleCombinations prop could be used, but the client sometimes wants them displayed but disabled instead of hiding them completely.

#### How to test it?

1. Link this branch on a workspace
2. Link the following PR from vtex.styleguide on the same workspace: https://github.com/vtex/styleguide/pull/1392
3. Add displayMode: "select" and disableUnavailableSelectOptions: true (or use my configuration) for sku-selector and check to see that the dropdown options that are actually "impossible" cases will now appear as disabled instead of enabled.  

`"sku-selector": {
    "props": {
      "visibility": "more-than-one",
      "hideImpossibleCombinations": false,
      "initialSelection": "empty",
      "displayMode": "select",
      "disableUnavailableSelectOptions": true
    }
  }`

https://user-images.githubusercontent.com/71461884/121042587-81c6ca80-c7bc-11eb-98dc-779ebcd833e7.mp4

hideImpossibleCombinations prop as alternative but the client doesn't want that behaviour so instead I had to go with this solution which is non-intrusive because of the extra prop that is by default set to false and only applies if the displayMode is "select" and if it's specified explicitly by the developer in store-theme (otherwise, in styleguide app, the disabled attribute will default to false and will have no impact)

#### Related to / Depends on

https://github.com/vtex/styleguide/pull/1392


react/components/SKUSelector/components/SelectVariationMode.tsx
```diff
@@ -24,6 +24,13 @@ function SelectVariationMode(props: VariationSelectModeProps) {
   const options = displayOptions.map(op => ({
     label: op.label,
     value: op.label,
+    disabled: op.disabled
+      ? op.impossible 
+        ? true
+        : !op.impossible && !op.available 
+        ? true 
+        : false
+      : false
```
victorhmp, line 33: Please break this into separate checks. It will greatly improve readability

victorhmp (changes requested):

react/components/SKUSelector/components/SelectVariationMode.tsx
razvanudream: Added a separate function for readability - you can see it more clearly now that if it's disabled && impossible then it's true, if it's not impossible and not available as well then true, for any other case it's false.

razvanudream:
Fixed conflicts again due to another PR going in master before this one (conflict was on CHANGELOG)

CHANGELOG.md
```diff
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
+### Added
+- Add `disableUnavailableSelectOptions` prop for `SKUSelector` in order to disable unavailable items for `displayMode` select
```
victorhmp, line 11: Just to make it easier, when you add an entry to the CHANGELOG, add it right below the `[Unreleased]` header, without an empty line. This will avoid conflicts :)

```suggestion
## [Unreleased]
### Added
- `disableUnavailableSelectOptions` prop to `SKUSelector` in order to disable unavailable items when `displayMode` is set to `select`.
```

victorhmp (approved):

docs/SKUSelector.md
```diff
@@ -41,6 +41,7 @@ The SKU Selector is a product details page block responsible for displaying ever
...
 | `visibleVariations`  | `array` | Specifies which product variations should be displayed in the product details page. Notice the following: if you declare a name that doesn't represent a real product variation or an empty array, no variations will be displayed. Check out more information regarding this prop structure below this table. | `undefined` |
 | `hideImpossibleCombinations`     | `boolean`   | Whether a product variation that leads to a impossible product combination should be clickable (`true`) or not (`false`). Notice that in both scenarios the variation will be displayed. However, when the prop is set as `false`, the variation is faded displayed (with less opacity). | `true`  |
+| `disableUnavailableSelectOptions`     | `boolean`   | Whether a product variation that leads to a impossible product combination should be displayed with the disabled attribute when `displayMode` is set to `select`. By default, all select options are without the disabled attribute set based on the type of variation | `false`  |
```
igorbrasileiro, line 44:
```suggestion
| `disableUnavailableSelectOptions`     | `boolean`   | Whether a product variation that leads to an impossible product combination should be displayed with the disabled attribute when `displayMode` is set to `select`. By default, all select options are without the disabled attribute set based on the type of variation | `false`  |
```

react/components/SKUSelector/components/SelectVariationMode.tsx
```diff
@@ -18,12 +18,28 @@ const messages = defineMessages({
...
+function isDisabled(
+  disabled: boolean,
+  impossible: boolean,
+  available: boolean
+) {
+  if (disabled) {
+    if (impossible) return true
+    if (!impossible && !available) return true
+
+    return false
```
igorbrasileiro, line 30: You don't need this return, the last return will be always false.
```suggestion
```

react/components/SKUSelector/components/SelectVariationMode.tsx
```diff
@@ -18,12 +18,28 @@ const messages = defineMessages({
   },
 })
 
+function isDisabled(
+  disabled: boolean,
+  impossible: boolean,
+  available: boolean
```
igorbrasileiro, line 24:
```suggestion
 { disabled, impossible, available }: IsDisableParams
```

```ts
interface IsDisableParams  {
  disabled: boolean,
  impossible: boolean,
  available: boolean
 }
```

igorbrasileiro (changes requested):
Congratz @razvanudream , nice job. I left some request changes to enhance the code.

razvanudream:
> Congratz @razvanudream , nice job. I left some request changes to enhance the code.

Thanks @igorbrasileiro - I made the changes you requested as it improves code readability. Cheers!

react/components/SKUSelector/components/SelectVariationMode.tsx
```diff
@@ -11,19 +11,35 @@ interface VariationSelectModeProps {
...
+  return false
+}
+
 function SelectVariationMode(props: VariationSelectModeProps) {
   const { intl, selectedItem, displayOptions } = props
 
   const options = displayOptions.map(op => ({
     label: op.label,
     value: op.label,
+    disabled: isDisabled({disabled: op.disabled, impossible: op.impossible, available: op.available}),
```
igorbrasileiro, line 42: Just a tip, isn't a request change.
```tsx
     .map({ disabled, impossible, available, label }) => ({
        label,
        value: label,
        disabled: isDisabled({ disabled, impossible, available }),
      })
```

igorbrasileiro: Can you lint this file?

igorbrasileiro (approved):

react/components/SKUSelector/components/SelectVariationMode.tsx
razvanudream: Ah yes, sorry about that, I ran eslint --fix on that file and found 2 warnings and 3 errors.  Should be fine now.

daniserejo (approved):

icazevedo (approved):

icazevedo:
Hey @razvanudream, could you please rebase this with master? I tried to do it, but I couldn't because I don't have the permissions required to push the rebased version to your repo.

icazevedo:
@allcontributors add @razvanudream for code

PR #958, opened 2021-07-02
Add the property name to SKU Selector dropdown

igorbrasileiro:
Fix a bug that the ProductSummarySKUSelector on mode select doesn't open the select dropdown due to that event is default prevented inside the onMouseDown function from the slider with multiples pages.

#### How to test it?
[Fixed Workspace](https://brasileirovtex--dacris.myvtex.com/marker-pentru-tabla-pilot-vboard-master-varf-rotund-2-3-mm/p?__bindingAddress=www.dacris.net/)
[Production workspace](https://dacris.myvtex.com/marker-pentru-tabla-pilot-vboard-master-varf-rotund-2-3-mm/p?__bindingAddress=www.dacris.net/)

#### Describe alternatives you've considered if any.
Considered use another HTML property, but the name seems good.

#### Related to / Depends on
https://github.com/vtex-apps/slider/pull/51

### DIFF
--- a/react/components/SKUSelector/components/SelectVariationMode.tsx
+++ b/react/components/SKUSelector/components/SelectVariationMode.tsx
@@ -41,7 +41,7 @@ function SelectVariationMode(props: VariationSelectModeProps) {
 
   return (
     <div className={styles.skuSelectorSelectContainer}>
-      {/* Warning: don't change this name property because it's used to check on Slider component, to avoid preventDefault of the event trigger by the dropdown */}
+      {/* Warning: don't change the name property. Its value is being checked in the Slider component to avoid a preventDefault call on the event triggered by the dropdown. */}
       <Dropdown
         name="product-summary-sku-selector"
         options={options}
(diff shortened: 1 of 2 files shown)


react/components/SKUSelector/components/SelectVariationMode.tsx
```diff
@@ -41,7 +41,9 @@ function SelectVariationMode(props: VariationSelectModeProps) {
 
   return (
     <div className={styles.skuSelectorSelectContainer}>
+      {/* Warning: don't change this name property because it's used to check on Slider component, to avoid preventDefault of the event trigger by the dropdown */}
```
victorhmp, line 44:
```suggestion
      {/* Warning: don't change the name property. Its value is being checked in the Slider component to avoid a preventDefault call on the event triggered by the dropdown. */}
```

victorhmp (approved):

icazevedo (approved):

PR #967, opened 2021-08-09
Dispatch set_loading_item if skuid is different

igorbrasileiro:
Prevent dispatch of the set_loading_item if SKU id is equal to the query. This avoids disabling the add-to-cart-button.
Related to https://github.com/vtex-apps/add-to-cart-button/pull/92 https://github.com/vtex-apps/product-context/pull/61

#### How to test it?

[Workspace-alss](https://brasprod--alssports.myvtex.com/peakre-food-mountain-berry-granola/p)
[workspace-melisseiras](https://brasileiro--melisseiras.myvtex.com/melissa-sun-rodeo-vermelho-33530/p)

### DIFF
--- a/react/components/SKUSelector/index.tsx
+++ b/react/components/SKUSelector/index.tsx
@@ -291,6 +291,8 @@ const SKUSelectorContainer: FC<Props> = ({
 
   const dispatch = useProductDispatch()
 
+  const getSkuId = () => query?.skuId
+
   const onSelectItem = useCallback(
     ({
       name: variationName,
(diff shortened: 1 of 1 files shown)


react/components/SKUSelector/index.tsx
```diff
@@ -291,6 +291,8 @@ const SKUSelectorContainer: FC<Props> = ({
 
   const dispatch = useProductDispatch()
 
+  const getSkuId = () => query?.skuId
```
icazevedo, line 294: I think this should be a dependency of the `onSelectedItem` `useCallback`. I think you'll need to wrap it on a `useCallback` of it's own

icazevedo (changes requested):

PR #968, opened 2021-08-09
PromoView event for Info-Card component

matheusmunarao:

A new feature in Info-Card component that shows analytics informations about it. This feature was implemented from store-image component and changed as needed. 

#### How to test it?

Check the page and visualizate the Info-Card component, then you could execute the dataLayer function in your own console to see their informations. Also you can configurate the analytics informations in /admin/cms/site-editor Info Card.

[Workspace](https://newinfocard--acctglobal.myvtex.com)

image
image

### DIFF
--- a/react/components/InfoCard/index.js
+++ b/react/components/InfoCard/index.js
@@ -10,7 +10,10 @@ import {
 import { formatIOMessage } from 'vtex.native-types'
 import { useCssHandles } from 'vtex.css-handles'
 import { useOnView } from 'vtex.on-view'
+
 import RichText from 'vtex.rich-text/index'
+import { usePixel } from 'vtex.pixel-manager'
+
 
 import CallToAction from './CallToAction'
 import LinkWrapper from './LinkWrapper'
@@ -173,20 +176,21 @@ const InfoCard = ({
 
   const subheadClasses = `${handles.infoCardSubhead} t-body mt6 c-on-base ${alignToken} mw-100`
 
+  const infoCardRef = useRef(null)
+  
+  const { push } = usePixel()
+
   const promotionEventData =
     analyticsProperties === 'provide'
       ? {
           id: promotionId,
           name: promotionName,
-          creative: formattedSrc,
           position: promotionPosition,
         }
       : undefined
    
-  const imageRef = useRef(null)
-
   useOnView({
-    ref: imageRef,
+    ref: infoCardRef,
     onView: () => {
       if (analyticsProperties === 'none') return
 
@@ -209,6 +213,7 @@ const InfoCard = ({
         style={containerStyle}
         data-testid="container"
         id={htmlId}
+        ref={infoCardRef}
         {...containerAttributes}
       >
         <div className={textContainerClasses}>
@@ -285,10 +290,6 @@ MemoizedInfoCard.propTypes = {
   // https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a
   linkTarget: oneOf(['_self', '_blank', '_parent', '_top']),
   callToActionLinkTarget: oneOf(['_self', '_blank', '_parent', '_top']),
-  analyticsProperties: bool,
-  promotionId: string,
-  promotionName: string,
-  promotionPosition: string,
 }
 
 MemoizedInfoCard.defaultProps = {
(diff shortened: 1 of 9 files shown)


react/components/InfoCard/index.js
```diff
@@ -168,6 +176,32 @@ const InfoCard = ({
 
   const subheadClasses = `${handles.infoCardSubhead} t-body mt6 c-on-base ${alignToken} mw-100`
 
+  const infoCardRef = useRef(null)
+  
+  const { push } = usePixel()
+
+  const promotionEventData =
```
igorbrasileiro, line 183: This seems to be used only inside the onView function, what do you think to put it there?

igorbrasileiro (approved):
Need approval from @vtex-apps/localization maybe.

react/components/InfoCard/index.js
matheusmunarao: Hi @igorbrasileiro! Thanks for review my PR.

Right! I changed the localization of promotionEventData to inside the useOnView. I'm not sure if this are your request too or only the promotionEventData, but hooks aren't must localize in the body of the component? I can't changed this to inside the useOnView.

react/components/InfoCard/index.js
```diff
@@ -95,6 +99,10 @@ const InfoCard = ({
   htmlId,
   textMode,
   linkTarget,
+  analyticsProperties = 'none',
+  promotionId,
+  promotionName,
+  promotionPosition,
```
victorhmp, line 105: Since these props would only be useful if `analyticsProperties` is set to something different than `none`, why not make `analyticsProperties` an object with these props? Then its default value would be `undefined`.

react/components/InfoCard/index.js
victorhmp: Hi @matheusmunarao!
I was about to leave the same suggestion as @igorbrasileiro 😅 .

I think what he meant is that this variable `promotionEventData` is only being used inside of the `onView` callback you're passing to the `useOnView` hook. Hence this variable doesn't need to be in the component's body but could be moved inside of the callback.

react/components/InfoCard/index.js
```diff
@@ -168,6 +176,32 @@ const InfoCard = ({
...
+          id: promotionId,
+          name: promotionName,
+          position: promotionPosition,
+        }
+      : undefined
+   
+  useOnView({
+    ref: infoCardRef,
+    onView: () => {
+      if (analyticsProperties === 'none') return
```
victorhmp, line 195: Very small nitpick, but we prefer to always write our `if`s as a block to make it easier to read 😄 

```suggestion
      if (analyticsProperties === 'none') {
        return
      }
```

victorhmp (changes requested):

camilavcoutinho (approved):

react/components/InfoCard/index.js
matheusmunarao: Hey @victorhmp, really sorry for the delay, I've had some big demands here.

As stated in the PR, this new feature was implemented from store-image component and in this component these props are declared in the same mode. I tried to change they as your request, but I couldn't do this without changing the admin,   noticing some problems in site-editor.

react/components/InfoCard/index.js
matheusmunarao: Done.

react/components/InfoCard/index.js
matheusmunarao: @victorhmp

igorbrasileiro: Add these properties on propTypes.

react/components/InfoCard/index.js
```diff
@@ -168,6 +176,36 @@ const InfoCard = ({
...
+
+      if (analyticsProperties === 'none') {
+        return
+      } else {
+        push({
+          event: 'promoView',
+          promotions: [promotionEventData],
+        })
+      }
+    },
```
igorbrasileiro, line 204: analyticsProperties can have which values? If it only has 'provide' and 'none,' you can improve your onView function.

```suggestion
    onView: () => {
      if (analyticsProperties === 'none') {
        return
      }

      const promotionEventData = {
        id: promotionId,
        name: promotionName,
        position: promotionPosition,
      }
        
        push({
          event: 'promoView',
          promotions: [promotionEventData],
        })
    },
```
Does this suggestion make sense for you?

igorbrasileiro (changes requested):

PR #981, opened 2021-09-28
update shipping simulator on change seller

wagnerlduarte:

The shipping simulator have a behavior to control when need to update the `ShippingTable` results. You can see this here: https://github.com/vtex-apps/store-components/blob/efebaee9e9e981849018d1f3d224f948ab7a521b/react/components/ShippingSimulator/Wrapper.js#L116

The problem was because the `isMountedRef.current` did not update when the product context was changed. 
You can see the problem below:
busca-frete-buybox

With this implementation, when the prop `key` changes, the `ShippingSimulator` is updated and set `isMountedRef.current` to `false`, as when the product page is started, and follow the normal flow of page. As shown below:
busca-frete-buybox-funcionando

#### How to test it?

Access product page using [buybox-context](https://github.com/vtex-apps/buybox-context) configured and use shipping simulator.

[Workspace](https://buyboxdemo--pedrocruzio.myvtex.com/luxury-purple-ballon/p)

### DIFF
--- a/react/components/ShippingSimulator/Wrapper.js
+++ b/react/components/ShippingSimulator/Wrapper.js
@@ -160,6 +160,7 @@ const ShippingSimulatorWithOrderForm = ({
 
   return (
     <BaseShippingSimulatorWrapper
+      key={seller}
       skuId={skuId}
       seller={seller}
       country={country}
@@ -217,6 +218,7 @@ const ShippingSimulatorWrapper = props => {
   ) {
     return (
       <BaseShippingSimulatorWrapper
+        key={sellerId}
         country={country}
         skuId={skuId}
         seller={sellerId}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
+### Changed
+- Added key on `BaseShippingSimulatorWrapper` to recreate component and load current shipping address when change seller.
+
 ## [3.153.0] - 2021-09-24
 
 ### Added


react/components/ShippingSimulator/Wrapper.js
```diff
@@ -160,6 +160,7 @@ const ShippingSimulatorWithOrderForm = ({
 
   return (
     <BaseShippingSimulatorWrapper
+      key={seller}
```
igorbrasileiro, line 163: This solution is tricky and solves the problem, but in my opinion, it can be solved with a better approach without destroying and remount the whole component tree.

The problem is, when changing the location code, should change the table of shipping price? If yes, why not create an effect (can be useEffect) to handle this property change?

igorbrasileiro (changes requested):

react/components/ShippingSimulator/Wrapper.js
wagnerlduarte: Apparently the component already is remounted, but the `useRef` control the `useEffect` execution and prevents the `handleCalculateShipping()` from being called.

https://github.com/vtex-apps/store-components/blob/efebaee9e9e981849018d1f3d224f948ab7a521b/react/components/ShippingSimulator/Wrapper.js#L115-L127

My problem happens when the `sellerId` changes. Currently when the `sellerId` changes, the Shipping Simulator don't call the `handleCalculateShipping` with this new value. So, the Seller Table component to be continued showing the old values, you know?
It's happens because my new component [buybox-context](https://github.com/vtex-apps/buybox-context) can change the seller as "winner" of buybox. And the shipping simulator impact diretly on this, because if a seller has a better price with Price + Shipping Price, the seller can be changed based on this new values.

I simulated your suggestion to compare the solutions using the Profiller of React Developer Tools, and apparently the execution cost was higher.
I put this code on `Wrapper.js`, as you can see bellow:

```
  useEffect(() => {
    if (isMountedRef.current) {
      isMountedRef.current = false
    }
  }, [seller])
```

https://github.com/vtex-apps/store-components/blob/efebaee9e9e981849018d1f3d224f948ab7a521b/react/components/ShippingSimulator/Wrapper.js#L113-L114

wagnerlduarte: Here they are exported profile that I used to compare the performance render of `BaseShippingSimulatorWrapper` component

[BaseShippingSimulatorWrapper.zip](https://github.com/vtex-apps/store-components/files/7261040/BaseShippingSimulatorWrapper.zip)

wagnerlduarte: Hey @igorbrasileiro, can you help me? 😄

PR #993, opened 2021-11-15
Allow displaying color name in sku

NAVODYA123:

This PR is added to facilitate to allow an option to display the name of the color for a color specification instead of the image. A prop named `showNameOfColor` was added. When a boolean input of `true` is passed to this prop, the  sku selector will show the color name as a label in a clickable button. This allows to have flexibility to choose between a text value representation of the color specification or an image representation. The default value of this prop is set to `false`

#### How to test it?

[Workspace](https://benspdp--benscycle.myvtex.com/sugino-75-track-crankset-crankset_sugino_sugino75_434/p?skuId=636804)

Before
before

After
after


NAVODYA123:
> Hi! I'm VTEX IO CI/CD Bot and I'll be helping you to publish your app! 🤖
> 
> Please select which version do you want to release:
> 
> * [x]  Patch (backwards-compatible bug fixes)
> * [ ]  Minor (backwards-compatible functionality)
> * [ ]  Major (incompatible API changes)
> 
> And then you just need to merge your PR when you are ready! There is **no need** to create a release commit/tag.
> 
> * [ ]  No thanks, I would rather do it manually 😞

PR #1035, opened 2022-06-23
Fix/slugified SKU

iago1501:

Slugify can't replace special characters.

Nowadays we are not removing the special characters (` and ’) of the SKU variation, and then, we are facing problems to slugify it

#### How to test it?

With the fix:

[Workspace](https://testerns--kissnewyork.myvtex.com/paleta-de-blush-ruby-kisses/p?skuId=367)

image

Without the fix:

[Workspace](https://withoutfix--kissnewyork.myvtex.com/paleta-de-blush-ruby-kisses/p?skuId=367)

image

#### Related to / Depends on

Zendesk #593195

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 ### Added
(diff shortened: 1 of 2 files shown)


danzanzini (approved):
Nice fix 👏

PR #1041, opened 2022-07-08
refactor query fetch and add position data

spclu6:

This pr tries to solve the following:

- Uses the hook useDevice of the app 'vtex.device-detector' which returns more accurately what type of device the user is on than the one currently in use.

- It avoids the Graphql query to be performed every time the component is rendered thus avoiding to make calls before it loads the minimum information to obtain the personalized images.

- It refactors the way in which the latitude and longitude of the user can be obtained extracting that logic in a custom hook so that the Image component is cleaner.

#### How to test it?

[Workspace](https://imageprotocol--spclusellas.myvtex.com/?__bindingAddress=spclusellas.myvtex.com/)

An alternative that is being explored is to directly obtain the user's location through the ip in the backend and thus avoid waiting time on the client.

### DIFF
--- a/react/components/InfoCard/index.js
+++ b/react/components/InfoCard/index.js
@@ -1,15 +1,15 @@
 import classNames from 'classnames'
 import PropTypes, { bool, string, oneOf } from 'prop-types'
-import React, { memo } from 'react'
+import React, { memo, useEffect } from 'react'
 import { values } from 'ramda'
 import { injectIntl } from 'react-intl'
-import {
-  useRuntime,
-  useExperimentalLazyImagesContext,
-} from 'vtex.render-runtime'
+import { useLazyQuery } from 'react-apollo'
+import { useExperimentalLazyImagesContext } from 'vtex.render-runtime'
 import { formatIOMessage } from 'vtex.native-types'
 import { useCssHandles } from 'vtex.css-handles'
 import RichText from 'vtex.rich-text/index'
+import { useDevice } from 'vtex.device-detector'
+import { useRenderSession } from 'vtex.session-client'
 
 import CallToAction from './CallToAction'
 import LinkWrapper from './LinkWrapper'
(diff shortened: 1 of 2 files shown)


react/components/InfoCard/index.js
```diff
@@ -141,32 +136,40 @@ const InfoCard = ({
...
-  if (!error && !loading && data && data.getImage && data.getImage.url !== null && data.getImage.urlMobile !== null && imageProtocolId !== '') {
-    console.log('getImage InfoCard: ',data.getImage)
-    imageHref = formatIOMessage({ id: data.getImage.hrefImg , intl })
-    actionHref = formatIOMessage({ id: data.getImage.hrefImg , intl })
-    if(mobile){
-      finalImageUrl = formatIOMessage({ id: data.getImage.urlMobile, intl })
+  const [getPersonalizedImages, { data: imageData }] = useLazyQuery(
+    GET_IMAGE_PROTOCOL_IMAGES,
+    {
+      ssr: false,
```
tomymehdi, line 145: why are we forcing the client side rendering?

react/hooks/usePosition.ts
```diff
@@ -0,0 +1,46 @@
+import { useEffect, useState } from 'react'
+
+interface Coordinate {
```
tomymehdi, line 3: let's replace `Coordinate` for `LatLong`. `Coordinate` is confusing

react/hooks/usePosition.ts
```diff
@@ -0,0 +1,46 @@
...
+  latitude: number | undefined
+  longitude: number | undefined
+}
+
+interface Position {
```
tomymehdi, line 8: Maybe something like `LatLongRecord`, `LatLongHistory` or `LatLongTime` is a better fit here

tomymehdi (changes requested):

LeslieJRP (approved):
changes reviewed

PR #1043, opened 2022-07-14
Add 'colours' to possibleValues

simmbiote:
For our project we needed to change "colour" to "colours" for a number of reasons. This has caused the colour sku selector to stop working. Ideally we do not want to create a custom component just for this reason.

The colour Sku selector is no longer showing the `thumbnailImage`  on our sites after needing to change the attribute from "colour" to "colours"
image

#### How to test it?

[Workspace](Link goes here!)

### DIFF
--- a/react/components/SKUSelector/utils/index.ts
+++ b/react/components/SKUSelector/utils/index.ts
@@ -57,6 +57,7 @@ export const isColor = (variation: string) => {
     'cor',
     'color',
     'colour',
+    'colours',
     'colore',
     'farbe',
     'couleur',


simmbiote:
> Beep boop 🤖
> 
> I noticed you didn't make any changes at the `docs/` folder
> 
> * [x]  There's nothing new to document 🤔
> * [ ]  I'll do it later 😞
> 
> In order to keep track, I'll create an issue if you decide now is not a good time
> 
> * [ ]  I just updated 🎉🎉

PR #1052, opened 2022-08-03
Added analytics push event on description show-more/show-less click

gabrieleghio:

By now, no analytics event is pushed when the user click on the show-more/show-less product description button.

#### How to test it?

You can test here or in any other product that have a long description in the site: [Workspace](https://descriptionpr--itwhirlpool.myvtex.com/lavastoviglie-da-incasso-whirlpool-colore-argento-grande-capienza-wic-3c33-f-869991604360/p)

For testing you can reach the workspace link and then press the **show-more** button:

image

Then if you look at the developer console and type **dataLayer** you can see that the event is correctly triggered:

image

You can also have a look at **enhancedEcommerceEvents.ts** file:

image

Obviously the event is fired both for **show-more** clicks and **show-less** ones and distinguished by the field ```action``` (```push({event: 'productDescription', product, action: newValue ? 'show-less' : 'show-more' })```) that can be **show-more** or **show-less**.

### DIFF
--- a/react/ProductDescription.tsx
+++ b/react/ProductDescription.tsx
@@ -82,7 +82,7 @@ function ProductDescription(props: PropsWithChildren<Props>) {
 
       <div className={`${handles.productDescriptionText} c-muted-1`}>
         {collapseContent ? (
-          <GradientCollapse collapseHeight={220}>
+          <GradientCollapse collapseHeight={220} product={product}>
             <SanitizedHTML
               content={description}
               allowedTags={allowedTags}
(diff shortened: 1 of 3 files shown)


gabrieleghio:
> mented [5 minutes ago](#issuecomment-1203643088)

react/ProductDescription.tsx
```diff
@@ -82,7 +82,7 @@ function ProductDescription(props: PropsWithChildren<Props>) {
 
       <div className={`${handles.productDescriptionText} c-muted-1`}>
         {collapseContent ? (
-          <GradientCollapse collapseHeight={220}>
+          <GradientCollapse collapseHeight={220} product={product}>
```
danzanzini, line 85: I don't think it's a good idea to attach the `GradientCollapse` component to the product. The component has a `onCollapsedChange` prop. You can send the push through it.

This way, you keep the product logic here

danzanzini (changes requested):
Thanks for the contribution. I've left comments on code :)

Issue #1056, opened 2022-10-12
[BUG] onInputChange not working correctly


brenogcota:
**Describe the bug**
[this code](https://github.com/vtex-apps/store-components/blob/master/react/components/SearchBar/AutocompleteInput.tsx#L237) not working correctly because InputSearch component from vtex style guide receive a prop called onChange, not onInputChange

To fix: 
```jsx
 <InputSearch
  ref={inputRef}
  size="large"
  value={value}
  {...restProps}
  onChange={onInputChange}
  error={Boolean(inputErrorMessage)}
  errorMessage={inputErrorMessage}
  onSubmit={onGoToSearchPage}
/>

```
**To Reproduce**

**Expected behavior**

**Screenshots**

**Desktop (please complete the following information):**

**Smartphone (please complete the following information):**

**Additional context**

PR #1057, opened 2022-10-12
fix: input change event

brenogcota:

This PR fix this [issue](https://github.com/vtex-apps/store-components/issues/1056) in search bar component

#### How to test it?

this pr just add an event trigger code snippet

[Workspace](Link goes here!)


brenogcota:
@victorhmp could you see this pr? :)

PR #1060, opened 2022-11-09
Fix/swiper destroyed

iago1501:

When changing the product context through the shelf and using productImage carousel, Swiper sometimes is destroyed due new render of the page, and then, when changing its context it breaks the page because Swiper is not mounted yet

#### How to test it?

Keep changing the shelf below the product image, after a while, the page is going to crash

[before](https://supportproduct--yoyojeans.myvtex.com/blusa-tiras-para-ni%C3%B1as-junior-con-nido-de-abeja-negro-32030209/p?skuId=20174)

[after](https://carolteste--yoyojeans.myvtex.com/camiseta-para-ninas-estampada-en-frente-amarillopastel-33055153/p?skuId=11787 )

image

#### Related to / Depends on

Zendesk #682994

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
+### Fixed
+- Check if the swiper is not destroyed before changing initial state, which could cause the whole product UI to crash
+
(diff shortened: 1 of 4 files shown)


eduardoformiga (approved):
From my tests using the `after` workspace, I didn't get the error while browsing between PDPs from the Shelf. 🎉 
Good job! 👏

PR #1062, opened 2022-11-17
Keep searched term inside search-bar

brendacdMoura:

Preserve search query in the search-bar.
(https://vtex-dev.atlassian.net/browse/IS-2675?atlOrigin=eyJpIjoiNjk3NjJmMzljZjZkNDg1YjhkNzE3OTg3ZDQ2Y2I5OTQiLCJwIjoiaiJ9)

#### How to test it?

[Workspace](https://brenda--oficinareserva.myvtex.com)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,12 +15,14 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ### Fixed
 - English, Portuguese and Italian translations.
 
+### Added
+- Preserve searched term in the search-bar.
+
 ## [3.163.4] - 2022-11-22
 ### Fixed
 - Update Product Name doc.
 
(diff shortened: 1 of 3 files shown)


thalytafabrine (commented):
lgtm, I'm just wondering if it wouldn't make sense for this to be a prop of the search bar, so the stores would have the option to keep the term or keep it without term as it currently is

brendacdMoura:
> lgtm, I'm just wondering if it wouldn't make sense for this to be a prop of the search bar, so the stores would have the option to keep the term or keep it without term as it currently is

I believe it is interesting to keep the searched term because it is a way to help the user to refine the query faster, Alex has said that this is the default in nearly every search engine too.

hiagolcm (approved):

PR #1103, opened 2024-02-22
Performance TBT improvement rendering videos only it's selected

arturmagalhaesjr:

TBT when there are videos from Youtube or Vimeo.

Solution:

Only the target video is selected, and then it's rendered.

#### How to test it?

https://www.raamdecoratievantuiss.nl/cavendish-warme-steen-vouwgordijn-20258/p?workspace=labprod3

### DIFF
--- a/react/components/ProductImages/components/Carousel/index.js
+++ b/react/components/ProductImages/components/Carousel/index.js
@@ -190,7 +190,7 @@ class Carousel extends Component {
             id={i}
           />
         ) : (
-          <></>
+          <div />
         )
 
       default:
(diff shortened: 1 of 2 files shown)


wender (approved):

react/components/ProductImages/components/Carousel/index.js
```diff
@@ -183,12 +183,14 @@ class Carousel extends Component {
 
       case 'video':
         return (
-          <Video
-            url={slide.src}
-            setThumb={this.setVideoThumb(i)}
-            playing={i === this.state.activeIndex}
-            id={i}
-          />
+          i === this.state.activeIndex && (
```
icazevedo, line 186: I would render an empty div or a similar object in case this isn't true. This prevents the number of items in the carousel changing unexpectedly.

icazevedo (approved):

PR #1112, opened 2024-08-07
Add img props to infoCard component

diegoleone:

Currently we do not support `fetchpriority` & `preload` props for images within the native infoCard component. Therefore, I am adding those props, to allow better management of those assets. The implementation follows the example used in the vtex.store-image native component.
https://github.com/vtex-apps/store-image/blob/master/react/Image.tsx

#### How to test it?

Visit the Dunnes store in this workspace: https://preloadimg--dunnesstorespreprod.myvtex.com/

Screenshot 2024-08-07 at 12 36 11


beatrizmaselli:
@diegoleone thanks for your contribution! Could you please update the docs folder with this new prop for the infoCard and add a changelog for this change? Thanks ✨

react/components/InfoCard/index.js
```diff
@@ -97,6 +97,8 @@ const InfoCard = ({
   htmlId,
   textMode,
   linkTarget,
+  fetchpriority,
+  preload
```
beatrizmaselli, line 101: Please add the comma here or run a `yarn lint --fix`

beatrizmaselli (approved):
LGTM

react/components/InfoCard/index.js
```diff
@@ -290,6 +298,8 @@ MemoizedInfoCard.defaultProps = {
   textMode: textModeTypes.TEXT_MODE_HTML.value,
   linkTarget: '_self',
   callToActionLinkTarget: '_self',
+  fetchpriority: 'auto',
+  preload: '',
```
iago1501, line 302: Since this is a boolean, this prop should be false and not line in isFullModeStyle ''

iago1501 (changes requested):
Just a type fix

react/components/InfoCard/index.js
iago1501, line 302:
```suggestion
  preload: false,
```

diegoleone:
@iago1501 Thank you. All done.

iago1501 (approved):
LGTM

PR #1124, opened 2024-10-21
Update: SKU Selector logic for default 

beatrizmaselli:
**What problem is this solving?**
The SKU selector, when there were multiple SKUs with the same specifications, was always selecting the first SKU without considering the product’s sale feasibility. This issue could lead to displaying an SKU that is not actually available for purchase, affecting customer experience and conversions.

**How to test it?**
Ensure that you have multiple SKUs with identical specifications but differing in sale feasibility. Verify that the SKU selector now prioritizes SKUs that are available for sale when displaying the SKU options in the context.

[Workspace](https://selector--dunnesstores.myvtex.com/regular-fit-crew-neck-t-shirt-4541124dkna/p?skuId=20918)

**Screenshots or example usage:**

Before adjustment:
image

After adjustment:
image

**Describe alternatives you've considered, if any.**
No significant alternatives were considered, as the focus was on improving SKU selection based on sale feasibility.

**How does this PR make you feel? [:link:](http://giphy.com/)**


beatrizmaselli (commented):
LGTM

PR #1128, opened 2025-01-22
[SFS-2085] refactor: sortVariationsByLabel logic when not numbers and redability

iago1501:

Initially, the condition `!Number.isNaN(option.label)` was used to check if option.label is a number. However, this approach had a flaw. When `option.label` is not a number, `Number.isNaN(option.label)` returns false, and the ! operator inverts this to true. As a result, any non-numeric or invalid value would be incorrectly considered "non-NaN", causing the every test to return true falsely.

To address this issue, the logic was revised to [!Number.isNaN(Number(option.label))](https://github.com/airbnb/javascript#standard-library--isnan). This change ensures that only numeric values are considered valid, preventing incorrect results in the every test."

#### How to test it?

[Before](https://hanesdev--hanes.myvtex.com/0mhb005/p?skuId=180)

image

[After](https://iespinoza--hanes.myvtex.com/0mhb005/p?skuId=180)

image

#### Related to / Depends on

It closes the PR: https://github.com/vtex-apps/store-components/pull/1127

### DIFF
--- a/react/components/SKUSelector/components/SKUSelector.tsx
+++ b/react/components/SKUSelector/components/SKUSelector.tsx
@@ -46,6 +46,7 @@ function getShowValueForVariation(
   )
 }
 
+
 interface Props {
   seeMoreLabel: string
   maxItems: number
@@ -82,6 +83,18 @@ function isSkuAvailable(item?: SelectorProductItem) {
   }
 
   return seller.commertialOffer?.AvailableQuantity > 0
+}
+
+function sortOptionsNumerically(options: DisplayOption[]) {
+  return options.slice().sort((a, b) => parseFloat(a.label) - parseFloat(b.label));
+}
+
+function sortOptionsAlphabetically(options: DisplayOption[]) {
+  return options.slice().sort((a, b) => {
+    if (a.label < b.label) return -1;
+    if (a.label > b.label) return 1;
+    return 0;
+  });
 }
 
 const showItemAsAvailable = ({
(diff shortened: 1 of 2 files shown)


react/components/SKUSelector/components/SKUSelector.tsx
```diff
@@ -281,17 +294,15 @@ const variationNameToDisplayVariation =
...
-        if (allNumbers) {
-          return a.label - b.label
-        }
+      const sortedOptions = areAllNumbers
+        ? sortOptionsNumerically([...options])
+        : sortOptionsAlphabetically([...options])
 
-        return a.label < b.label ? -1 : a.label > b.label ? 1 : 0
-      })
+      options = [...sortedOptions]
```
vsseixaso, line 305: Here the code recreate the entire array 3 times. Less performant, it is possible to maintain the previous idea of ​​sorting directly in the original array:
```
if (areAllNumbers) {
  sortOptionsNumerically(options)
} else {
  sortOptionsAlphabetically(options)
}
```

vsseixaso (changes requested):

react/components/SKUSelector/components/SKUSelector.tsx
iago1501: The first approach was chosen because it adopts principles of immutability and code clarity. By using the spread operator ([...]) to create copies of the arrays before applying the sort functions, we guarantee that the original options array remains unchanged. With immutability principle we can prevent unwanted side effects and make it easier to track changes in state, although looking less performatic due to the copy of the arrays, changing directly the object might cause unexpected re-renders in the application

vsseixaso: Due to the lack of clarity of where it is used and the side effects, it is understandable.

vsseixaso (approved):

PR #1154, opened 2026-09-01
feat: forward priceToken from BuyButton to the cart

monteirogc:
## Summary

Legacy (v1) add to cart chain. `store-components` keeps its own `mapCatalogItemToCart` — a mirror of what `add-to-cart-button` does on the v2 chain — and that is where the search-signed price enters this path.

Three propagation points:

1. `BuyButton/Wrapper.js` — `mapCatalogItemToCart` reads `priceToken` off the selected seller's commercial offer
2. `BuyButton/index.js` — `skuItemToMinicartItem`, for the optimistic minicart item
3. `BuyButton/index.js` — variables of the direct `addItem` mutation, used when there is no local state

Without the first, the other two would have nothing to carry.

## No schema change needed

The `MinicartItem` referenced by the `@client` mutations is a phantom type: the minicart's local state registers only `resolvers` and `initialState`, no `typeDefs`, and stores the cart as a JSON string. The token rides along as a plain object field.

## Ordering

Draft on purpose. This can only merge after `store-graphql#699` publishes — before that, `priceToken` is not on `OrderFormItemInput` and GraphQL rejects the variable during coercion.

The other half of this chain is the minicart PR, which stops the token from being dropped by the field whitelist before the server sync.

## Related

[STR-1229](https://vtex-dev.atlassian.net/browse/STR-1229), part of [STR-1132](https://vtex-dev.atlassian.net/browse/STR-1132).

## Test plan

- [x] `yarn test BuyButton` — 10 passing, no regressions
- [ ] v1 chain end to end after `store-graphql#699` is published
- [ ] Item without token: payload identical to today

Made with [Cursor](https://cursor.com)

[STR-1229]: https://vtex-dev.atlassian.net/browse/STR-1229?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


react/components/BuyButton/index.js
```diff
@@ -44,6 +44,7 @@ const skuItemToMinicartItem = item => {
     seller: item.seller,
     options: item.options,
     quantity: item.quantity,
+    ...(item.priceToken ? { priceToken: item.priceToken } : {}),
```
mendescamara, line 47: Is this check necessary?

Doesn't simply adding the value to the routine handle cases where it is null or undefined?

```
priceToken: item.priceToken,
```

monteirogc: It is necessary, and the reason is one layer down — in how the v1 minicart forwards the item.

Writing `priceToken: item.priceToken` unconditionally would not be equivalent, because the field comes back as **`null`**, not `undefined`. Checked it on two accounts whose Intelligent Search does not sign prices:

```
hmlfg   -> {"Price":15.03,"priceToken":null,"type":"null","keyExists":true}
vasari  -> {"Price":5.21, "priceToken":null,"type":"null","keyExists":true}
```

So the key would exist with a `null` value. That matters because the v1 minicart selects the fields it syncs with Ramda's `pick`, and Ramda keys off **existence**, not value:

```js
// ramda/src/pick.js
while (idx < names.length) {
  if (names[idx] in obj) {
    result[names[idx]] = obj[names[idx]];
  }
  idx += 1;
}
```

Ran it to be sure rather than assume:

```
key absent           -> {"id":"1"}
priceToken=undefined -> {"id":"1"}
priceToken=null      -> {"id":"1","priceToken":null}
priceToken="eyJ..."  -> {"id":"1","priceToken":"eyJ..."}
```

`undefined` disappears (`pick` copies it, then `JSON.stringify` drops it), but `null` survives all the way into the mutation variables. The net effect would be `priceToken: null` riding on every add-to-cart for every offer without a signed price — which, sampling 35 accounts earlier today, is roughly one account in ten.

It is not harmful today: the input field is a nullable `String`, so validation passes, and `hasPriceToken` in `store-graphql` requires a non-empty string, so the request still takes the `POST` route. But it stays harmless only because three separate behaviours happen to line up, and it contradicts the premise this change is built on — that a cart without a token sends byte-for-byte what it sends today. `priceToken: null` is not that.

The same applies to the mapping at line 181, in the `callOrderFormDirectly` branch, where the payload goes straight to the server rather than through the minicart.

I have tests for exactly this question, ready to push: seven cases on `mapCatalogItemToCart`, three of which assert `'priceToken' in item === false` for `null`, `undefined` and empty string — the assertion is on key absence, not on `toBeUndefined()`, precisely because `toBeUndefined()` would pass while the bug shipped. There is also one asserting that an item built with a token and one built without differ in nothing else, so this change cannot be blamed for a difference in any other cart item field.

Happy to drop the conditional if you would rather keep the mapping flat, but then the `pick` list in `minicart#352` would have to filter by value instead, and I would rather have the guard at the point where the field enters the item.

react/components/BuyButton/Wrapper.js
```diff
@@ -237,6 +237,9 @@ EnhancedBuyButton.mapCatalogItemToCart = function mapCatalogItemToCart({
         seller: selectedSeller.sellerId,
         price: selectedSeller.commertialOffer.Price,
         listPrice: selectedSeller.commertialOffer.ListPrice,
+        ...(selectedSeller.commertialOffer.priceToken
+          ? { priceToken: selectedSeller.commertialOffer.priceToken }
+          : {}),
```
vsseixaso, line 242: Esse valor vem do resultado da query de produto, que fica em cache enquanto a
aba estiver aberta, e o token vale 30 min. PDP aberta por mais tempo manda token
vencido.

Se o checkout rejeitar, o addItem cai no catch do addToCartAndFinish e o usuário
leva um add-failure. Clicar de novo não resolve: o skuItems vem da mesma query em
cache, então repete o mesmo token até dar reload. No isOneClickBuy é pior, o
setTimeout redireciona pro checkout com o item fora do carrinho.

Mesmo caso do minicart, confirmar se o token expirado é ignorado ou rejeitado.

react/components/BuyButton/Wrapper.js
vsseixaso, line 254: O item pode sair com priceToken e assemblyOptions juntos, e aí o preço da linha é
Price + sumAssembliesPrice, diferente do que o token assinou. O checkout compara
o preço final com a claim ou usa o token só como fallback do preço base?

mendescamara (approved):
I missed having tests covering these changes.

react/components/BuyButton/Wrapper.js
monteirogc: Checkout ignores an expired token, it does not reject it. So neither the add-failure nor the one-click case happens.


```
A) no token at all         -> ok, 1 line, sellingPrice=52141
B) expired token (271 min) -> ok, 1 line, sellingPrice=52141, no error, no messages
C) corrupted signature     -> ok, 1 line, sellingPrice=52141, no error
```

react/components/BuyButton/Wrapper.js
monteirogc: The token is used as the price of that specific offer, not compared against the final line total — and in this flow it never travels on a request that carries assemblies.

vsseixaso (approved):

PR #88, opened 2018-06-07
Internationalization of the Share schema.

estacioneto:
#### What is the purpose of this pull request?
Add internationalization to the Share schema.

#### What problem is this solving?
The schema was English only and hardcoded.

#### How should this be manually tested?

[Access the workspace](https://estacio--storecomponents.myvtex.com/nintendo-switch/p) and see the schema

#### Screenshots or example usage

image

#### Types of changes
- [ ] Bug fix (a non-breaking change which fixes an issue)
- [x] New feature (a non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Requires change to documentation, which has been updated accordingly.

PR #1157, opened 2026-09-15
fix: keep deliberately cleared variations cleared on sku resync

iago1501:

The resync effect in `SKUSelector` re-derives the whole selection from the resolved SKU whenever `skuSelected` changes. With `initialSelection` defaulting to `complete`, that silently re-picks a variation the shopper had just cleared.

Reproducing it on `classic-shoes` (a product whose colours have more than one size):

1. Land on the PDP with White / 41 selected.
2. Click `41` to clear the size.
3. Click another colour.

The resolved SKU changes, the effect runs, and the size comes back selected — a size the shopper never chose. Because `areAllVariationsSelected` is computed from that same local selection, add-to-cart becomes enabled for it.

Note that `uniqueOptionToSelect` is *not* what fills the size here: every colour in the repro has two available sizes, so there is no unique option to resolve. It is the resync that re-derives it.

This PR tracks which variations were cleared on purpose and keeps them cleared across the resync, while still letting an external SKU change (URL navigation, quickview → PLP sync) update the rest of the selection.

#### How to test it?

[Workspace](https://storetheme.vtex.com/classic-shoes/p?workspace=qareview)

Both this branch and [vtex-apps/product-context#88](https://github.com/vtex-apps/product-context/pull/88) are linked there.

1. Open the PDP and confirm White / 41 is selected.
2. Click `41` to clear the size.
3. Click the Green swatch.

Expected: `Color: Green`, **no** size selected, and add-to-cart blocked. On `master` the size comes back selected.

The regression test added here covers exactly this: removing the fix while keeping the test makes it fail on the assertion after the resync, with the size re-selected.

Verified live by reading the product context straight off the React Fiber tree and matching the gallery image id against each SKU's images, rather than eyeballing a screenshot:

| Scenario | `selectedItem` | pin | image matches item? | label |
| --- | --- | --- | --- | --- |
| Fresh load, no `skuId` | White `310124175` | `null` | yes | Color: White, Size 41 |
| Clear `41` → click Green | Green `310124227` | `310124227` | yes | Color: Green, no size |
| Then pick Size `40` | `35` | `null` (cleared) | yes | add-to-cart enabled |

**Resyncing the label to the resolved item instead.** Coherent, but it visibly undoes the shopper's colour click, as if the click had been rejected.

**Redirecting to the candidate SKU when the selection is incomplete**, instead of clearing `skuId`. Better UX on paper, but it puts a concrete `skuId` back in the query, which sends `getNewSelectedVariations` down the `hasSkuInQuery` branch and auto-completes the selection — the very bug this PR fixes, just reached from the other side.

#### Related to / Depends on

Enables [vtex-apps/product-context#88](https://github.com/vtex-apps/product-context/pull/88), which makes the incomplete-selection fallback follow the colour the shopper clicked. Without this PR, the resolved item moving to that colour would trip the auto-complete described above.

This PR is independently correct and can land on its own.


react/components/SKUSelector/index.tsx
```diff
@@ -275,11 +315,38 @@ const SKUSelectorContainer: FC<Props> = ({
...
+    if (!isSameProduct(previousItemsRef.current, parsedItems)) {
+      clearedVariations.current = new Set()
+    }
+
+    previousItemsRef.current = parsedItems
+
     setSelectedVariations(
-      getNewSelectedVariations(query, skuSelected, variations, initialSelection)
+      keepClearedVariations(
+        getNewSelectedVariations(
```
mendescamara, line 341: This only forces `null` for names already in `clearedVariations`. Everything else is rewritten from `skuSelected` via `getNewSelectedVariations`.

That is enough to stop Size from auto-completing, but a colour the shopper just clicked is **not** in the Set, so it comes back from the resolved SKU. If this ships without [product-context#88](https://github.com/vtex-apps/product-context/pull/88) (fallback following the pinned colour) and `skuSelected` changes identity onto the catalog fallback, the colour click is undone while Size stays empty — gallery, selector and add-to-cart disagree.

Please treat the two PRs as one PDP rollout (or confirm the incomplete-selection flow on a workspace that does **not** include #88).

react/components/SKUSelector/index.tsx
```diff
@@ -263,6 +295,14 @@ const SKUSelectorContainer: FC<Props> = ({
   const responsiveDisplayMode = useResponsiveValue(displayMode)
 
   const parsedItems = useMemo(() => skuItems.map(parseSku), [skuItems])
+  /* Content-derived key, not the array reference: callers that pass a
+   * freshly-built `skuItems` array on every render (e.g. via `.filter()`)
+   * would otherwise make the resync effect below re-run on every render. */
+  const itemsKey = useMemo(
+    () => parsedItems.map(item => item.itemId).join('|'),
+    [parsedItems]
```
mendescamara, line 308: Content key is the right idea (avoids resync on an unstable `skuItems` reference), but this is order-sensitive: the same item set in a different order produces a different `itemsKey` and re-runs `getNewSelectedVariations`.

With `initialSelection="empty"` that can wipe a partial selection that was never added to `clearedVariations`.

Sort a copy of the ids before joining, e.g. `[...parsedItems.map(item => item.itemId)].sort().join('|')`. Worth a test that inverted `skuItems` order does not change `selectedVariations`.

iago1501: Good catch, fixed — the key is now `[...ids].sort(...).join('|')`.

I confirmed the order really does flap in practice: reading `previousItemsRef` off the live component between two renders on `classic-shoes` gave `["35","37","310124175","310124227"]` and then `["310124175","310124227","35","37"]` for the same four items.

Worth noting for the record that the reproduction is narrower than it looks: going through `Wrapper`, `variations` is derived from `skuItems` via a `useMemo` keyed on that array's reference, so any reorder already produces a new `variations` object and re-runs the effect regardless of `itemsKey`. The sort matters for callers that pass `variations` and `skuItems` as independent props (the exported container's actual contract), which is what the regression test added here exercises — it renders the container directly with a stable `variations` reference and only reorders `skuItems`, and it fails without the sort.

react/components/SKUSelector/index.tsx
iago1501: I went to reproduce this on a workspace without [product-context#88](https://github.com/vtex-apps/product-context/pull/88), and the conclusion is that you're right to flag the flow — but the mechanism is different from "the colour comes back from the resolved SKU", and it's worse.

What actually happens is that the whole selector **remounts**. Starting on an item that isn't the catalog fallback (`?skuId=37`, sold out), clearing Size and clicking an ambiguous colour, the `.skuSelectorContainer` DOM node is a different node afterwards, and both `useRef`s (`clearedVariations`, `previousItemsRef`) are fresh objects. So nothing here gets a chance to protect anything: `keepClearedVariations` only runs from the resync effect, and `useEffectSkipMount` skips it on mount by design.

The trigger is the theme, not either app. `store-theme` has, in `store/blocks/pdp/product.jsonc`:

```jsonc
"condition-layout.product#availability": {
  "props": {
    "conditions": [{ "subject": "isProductAvailable" }],
    "Then": "flex-layout.row#product-main",
    "Else": "flex-layout.row#product-availability"
  }
}
```

Both branches list `sku-selector`, but as separate block trees, so when `isProductAvailable` flips React unmounts one and mounts the other. Evidence:

- The `flexColChild--info-availability` class (the `Else` branch) is present before the interaction, since item `37` is sold out.
- The `vtex-render__container-id-availability` node survives while its entire subtree (selector, brand, gallery) is replaced; the breadcrumb, a sibling outside the conditional, survives.
- It's the availability flip, not the SKU or query change: `310124175` → `310124227` (both available) does **not** remount, while `37` (sold out) → `310124175` (available) does.
- It reproduces on vanilla production, with neither PR linked, so it predates both.

On your actual ask — treating the two as one rollout. Splitting them is less dangerous than it first looks, because `ProductContextProvider` sits above that conditional and **survives** the remount, so everything the KI fix depends on (clearing the pin when it disagrees, and the fallback following the clicked colour) is unaffected. What the remount does defeat is this PR's protection, in that one availability-flipping transition: the selection auto-completes again and add-to-cart opens up for a size the shopper cleared. That's the same in both directions of the rollout, so it isn't an argument for ordering them — it's a separate, pre-existing gap.

I've documented it as a known limitation in the CHANGELOG and next to the ref itself rather than trying to work around it here. Closing it properly would mean holding that tracking in `product-context` state (which survives) and consulting it from the `useState` initializer too, since a remount skips the effect — a change on both sides. Happy to do that if you'd rather not ship with the gap.

mendescamara (approved):

